PR #62, opened 2025-05-16
Implement unit tests

leonardostsouza:


src/utils/functions.js
```diff
@@ -2,6 +2,8 @@ import b4a from 'b4a';
 import sodium from 'sodium-native';
 import {peer} from "hyperdht/lib/messages.js";
 
+// TODO: (?) Should we allow 0x prefix?
+// TODO: (?) Should an mepty string be considered valid?
```
b-garbacz: // TODO: (?) Should an mepty string be considered valid? - No

src/utils/functions.js
```diff
@@ -2,6 +2,8 @@ import b4a from 'b4a';
 import sodium from 'sodium-native';
 import {peer} from "hyperdht/lib/messages.js";
 
+// TODO: (?) Should we allow 0x prefix?
```
b-garbacz: as you mentioned in the past - lets adopt bitcoin standard which doesn't accept 0x, however we should consider a case - do we accept uppercases? Are we going only to accept lowercase hexstring?

src/utils/msgUtils.js
```diff
@@ -16,11 +16,27 @@ class MsgUtils {
         return buf;
     }
 
+    static #checkAssembleMessageBaseParams(wallet, keyParam) {
+        return !((!wallet || !keyParam) || 
+            (typeof wallet !== 'object') || 
+            (typeof keyParam !== 'string') || 
+            (keyParam.length !== 64) ||
```
b-garbacz, line 23: This is not possible to write a schema in check js for it?

test/functions/isHexString.test.js
```diff
@@ -0,0 +1,10 @@
+import test from 'brittle';
+import { isHexString } from '../../src/utils/functions.js';
+
+test('isHexString', (t) => {
+    // t.ok(isHexString('0x1234567890abcdef'), 'Valid hex string should return true'); // Deactivated. See TODO in functions.js
```
b-garbacz, line 5: should return false

test/functions/isHexString.test.js
```diff
@@ -0,0 +1,10 @@
...
+    t.ok(isHexString('1234567890abcdef'), 'Valid hex string should return true');
+    t.ok(isHexString('1234567890xyz') === false, 'Invalid hex string should return false');
+    t.ok(isHexString('123456789') === false, 'Invalid size hex string should return false');
+    // t.ok(isHexString('') === false, 'Empty string should return false'); // Deactivated. See TODO in functions.js
```
b-garbacz, line 9: should return false

test/functions/isHexString.test.js
```diff
@@ -0,0 +1,10 @@
...
+});
```
b-garbacz, line 10: todo: let's discuss about uppercase characters

b-garbacz (changes requested):

b-garbacz (approved):

src/utils/functions.js
leonardostsouza: The current implementation already accepts uppercase

PR #105, opened 2025-06-27
Refactor directory structure

b-garbacz:


Whitelist/pubkeys.csv
leonardostsouza, line 1: This file should be in the .gitignore in the future

package.json
```diff
@@ -32,7 +32,7 @@
     "protomux-wakeup": "2.4.0",
     "readline": "npm:bare-node-readline",
     "ready-resource": "1.1.2",
-    "trac-wallet": "git+https://github.com/Trac-Systems/ed25519-key-generator.git#trac-wallet-r2",
+    "trac-wallet": "0.0.43",
```
leonardostsouza, line 35: Are you sure you want to revert to the unmodified version?

leonardostsouza (approved):
Overall, just pay attention to the comments, but I wont block it. Approved

Whitelist/pubkeys.csv
b-garbacz: Yeah, I will create a ticket for this

package.json
b-garbacz: Yes, we will change it on Monday

PR #114, opened 2025-07-01
Implement encoders and decoders for admin and node entries

leonardostsouza:


src/core/state/encodings.js
```diff
@@ -0,0 +1,143 @@
...
+const WRITING_KEY_SIZE = 32;
+const TRAC_PUB_KEY_SIZE = 32;
+const TRAC_ADDRESS_SIZE = 1 + TRAC_PUB_KEY_SIZE; // TODO: This will change in the future
+const ADMIN_ENTRY_SIZE = 1 + TRAC_ADDRESS_SIZE + WRITING_KEY_SIZE;
+const NODE_ENTRY_SIZE = WRITING_KEY_SIZE + 1;
+
+// Role masks
+// TODO: Consider implementing a role mask for Admin
+const WRITER_MASK = 0x1;
+const INDEXER_MASK = 0x2;
```
b-garbacz, line 13: I think you can move all of these constants into the constants.js

src/core/state/encodings.js
```diff
@@ -0,0 +1,143 @@
...
+    if (isBufferValid(nodeEntry, NODE_ENTRY_SIZE)) {
+        nodeEntry[0] = (isWriter ? WRITER_MASK : 0x0) | (isIndexer ? INDEXER_MASK : 0x0);
+    }
+    return nodeEntry;
+}
+
+// ------------ TRANSACTIONS ------------ //
+export function encodeTransaction(tx) {
+    // TODO
+    // [OP(1)][TX_HASH(??)][IS(??)][W(32)][I(??)][IPK(??)][CH(??)][IN(??)][BS(32)][MBS(32)][WS(??)][WP(??)][WN(32)]
```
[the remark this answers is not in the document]
b-garbacz: So it means, probably we won't need it

test/all.test.js
```diff
@@ -5,6 +5,7 @@ import { default as test } from 'brittle';
 async function runTests() {
   test.pause();
 
+  // await import('./encodings/encodings.test.js'); // it should work
```
b-garbacz, line 8: Hmn. I would like to suggest to change name of this module to ApplyOperationEncodings.test.js? What do you think about it?

test/encodings/encodings.test.js
```diff
@@ -0,0 +1,94 @@
...
+    encodeAdminEntry,
+    decodeAdminEntry,
+    encodeNodeEntry,
+    decodeNodeEntry,
+    setNodeEntryRole
+} from '../../src/core/state/encodings.js';
+
+const WRITING_KEY_SIZE = 32;
+const TRAC_PUB_KEY_SIZE = 32;
+const TRAC_ADDRESS_SIZE = 1 + TRAC_PUB_KEY_SIZE;
```
b-garbacz, line 13: if these sizes will be moved into constants.js then you can just import them

src/core/state/encodings.js
```diff
@@ -0,0 +1,143 @@
...
+// ------------ NODE ENTRY ------------ //
+
+/**
+ * Encodes a node entry as a buffer containing the role mask and writing key.
+ * The first byte is a bitmask indicating writer and indexer roles.
+ * The remaining bytes are the node's writing key.
+ *
+ * NODE_ENTRY = [NODE_ROLE_MASK(1)][WRITING_KEY(32)]
+ * NODE_ROLE_MASK:
+ *   - WRITER = 0x1 (bit 0)
```
b-garbacz, line 78: what if both are turned on?

src/core/state/encodings.js
leonardostsouza: Better not. They are meant to be used locally only.

test/encodings/encodings.test.js
leonardostsouza: I don't want to expose them. We should call only the exported functions. There is no use accessing these values outside the encoders and decoders (only for testing)

src/core/state/encodings.js
leonardostsouza: This is a bit mask. If a node is both an Indexer and a Writer, the value in this field will be `0x3`

test/encodings/encodings.test.js
b-garbacz: ok

b-garbacz (approved):
looks fine

PR #122, opened 2025-07-22
Fix preTx and postTx for binary data format

b-garbacz:


b-garbacz:
@leonardotc  - I know that it's after 10 pm. However I had programming flow to fix pre and post with optimalized Tx. Clean up is left and we can merge everything to R2

b-garbacz:
We are already saving 64 bytes, which is good result imo.

leonardostsouza (approved):

PR #126, opened 2025-07-23
124 clean up network module

b-garbacz:


src/core/network/messaging/validators/AdminResponse.js
```diff
@@ -0,0 +1,57 @@
...
+    async validate(message, channelString) {
+        if (!this.validatePayload(message)) return false;
+        if (!this.validateIssuerPublicKey(message)) return false;
+        if (!this.validateTimestamp(message)) return false;
+        if (!await this.validateAdminData(message)) return false;
+        if (!await this.validateAdminSignature(message)) return false;
+        if (!this.validateChannel(message, channelString)) return false;
+
+        return true;
+    }
```
leonardostsouza, line 20: It would be better to just use one if with the `||` operator. It is cleaner than using so many returns.

src/core/network/messaging/validators/CustomNodeResponse.js
```diff
@@ -0,0 +1,44 @@
...
+    async validate(message, channelString) {
+        if (!this.validatePayload(message)) return false;
+        if (!this.validateIssuerPublicKey(message)) return false;
+        if (!this.validateTimestamp(message)) return false;
+        if (!await this.validateCustomNodeEntry(message)) return false;
+        if (!await this.validateCustomNodeSignature(message)) return false;
+        if (!this.validateChannel(message, channelString)) return false;
+
+        return true;
+    }
```
leonardostsouza, line 18: Same as previous comment

src/core/network/messaging/validators/ValidatorResponse.js
```diff
@@ -0,0 +1,57 @@
...
+        if (!this.validatePayload(message)) return false;
+        if (!this.validateIssuerPublicKey(message)) return false;
+        if (!this.validateTimestamp(message)) return false;
+        if (!await this.validateNodeEntry(message)) return false;
+        if (!await this.validateWritingKey(message)) return false;
+        if (!await this.validateValidatorSignature(message)) return false;
+        if (!this.validateChannel(message, channelString)) return false;
+
+        return true;
+    }
```
leonardostsouza, line 19: Same as previous comment

src/core/network/messaging/validators/base/baseResponse.js
```diff
@@ -0,0 +1,113 @@
...
+        const verified = this.#wallet.verify(signature, hash, customNodePublicKey);
+
+        if (!verified) {
+            console.error("Custom node response verification failed");
+            return false;
+        }
+
+        return true;
+    }
+}
```
leonardostsouza, line 111: All signature validation methods are basically the same function copy-pasted. It would be better to create a basic "signature validation" function and just invoke it with different arguments in the specialized functions

src/core/network/services/ValidatorObserverService.js
```diff
@@ -0,0 +1,84 @@
...
+        if (validatorAddress === address) return;
+
+        const validatorPubKey = Wallet.decodeBech32m(validatorAddress).toString('hex');
+        const validatorEntry = await this.state.getNodeEntry(validatorAddress);
+
+        if (this.#network.validator_stream !== null) return;
+        if (this.#network.validator !== null) return;
+        if (validatorEntry === null) return;
+        if (!validatorEntry.isWriter) return;
+        if (validatorEntry.isIndexer) return;
```
leonardostsouza, line 70: Again, it would be cleaner (and probably faster) to just use an `||` operator

src/core/network/messaging/validators/CustomNodeResponse.js
b-garbacz: Done

src/core/network/messaging/validators/ValidatorResponse.js
b-garbacz: Done

leonardostsouza (approved):

PR #127, opened 2025-07-24
Clean up State Module

leonardostsouza:


b-garbacz:
I will convert it to draft

src/core/state/utils/adminEntry.js
```diff
@@ -0,0 +1,66 @@
...
+ * Encodes an admin entry as a buffer containing the TRAC address and writing key.
+ * 
+ * The buffer format is: [TRAC_ADDRESS][WRITING_KEY(32)]
+ * Where TRAC_ADDRESS is a bech32m-encoded address without the HRP and separator.
+ *
+ * @param {Buffer} tracAddr - The TRAC address.
+ * @param {Buffer} wk - The admin's writing key buffer (must be 32 bytes).
+ * @returns {Buffer} The encoded admin entry buffer, or an empty buffer if input is invalid.
+ */
+export function encode(tracAddr, wk) {
```
b-garbacz, line 19: I was thinking about the name of this parameter: tracAddr. Maybe just address is enough. I think everyone probably knows it's a TRAC address.

src/core/state/utils/indexerEntry.js
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * @param {Buffer} indexerAddr - The indexer address to append (must be TRAC_ADDRESS_SIZE bytes).
+ * @param {Buffer|null} indexersEntry - The current indexers entry buffer, or null to create a new one.
+ * @returns {Buffer} The updated indexers entry buffer or an empty buffer in case something goes wrong.
+ */
+export function append(indexerAddr, indexersEntry = null) {
+    if (!isBufferValid(indexerAddr, TRAC_ADDRESS_SIZE)) {
+        return b4a.alloc(0);
+    }
+    // Append the indexer address to the database
```
b-garbacz, line 20: Typo: Add the indexer address to the entry.
This function is not responsible for modifying the state.

src/core/state/utils/indexerEntry.js
```diff
@@ -0,0 +1,105 @@
...
+    try {
```
b-garbacz, line 21: This is optimalization to discuss - instead of using concat we can allocate required size before. It would look like (case if entry exists):
// calculate count
 const currentCount = indexersEntry[0];
 const newCount = currentCount + 1;
 const updatedEntry = b4a.alloc(1 + newCount * TRAC_ADDRESS_SIZE);
 // set new count
 ...
 // copy previous state of indexers and copy again to append the new addres at the end. 
 
 This is only proposal

src/core/state/utils/indexerEntry.js
```diff
@@ -0,0 +1,105 @@
...
+ * The buffer format is: [count(1)][indexerAddr1][indexerAddr2]...[indexerAddrN]
+ * If the indexer address is not found or input is invalid, returns an empty buffer.
+ *
+ * @param {Buffer} indexerAddr - The indexer address to remove (must be TRAC_ADDRESS_SIZE bytes).
+ * @param {Buffer} indexersEntry - The current indexers entry buffer.
+ * @returns {Buffer} The updated indexers entry buffer with the address removed,
+ *                   or an empty buffer if the address is not found or input is invalid.
+ */
+export function remove(indexerAddr, indexersEntry) {
+    if (!b4a.isBuffer(indexersEntry) ||
```
b-garbacz, line 75: also here instead of using concat we can utilize copy with preallocated buffer (we know the size) and remove what we need

PR #145, opened 2025-08-06
142 implement ci

b-garbacz:


test/messageOperations/commonsStateMessageOperationsTest.js
```diff
@@ -79,16 +79,21 @@ export async function messageOperationsEkoTest(t, fnName, assembler, wallet, wri
...
-            async () => await assembler(
-                wallet,
-                b4a.from("1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdeg", 'hex')),
-            errorMessageIncludes('Writer key must be a 32 length buffer')
-        );
-
-    });
+    //
+    // Test removed as b4a.from() with 'hex' encoding will throw on invalid hex characters,
+    // making this test redundant. The validation is handled by the b4a library itself.
```
leonardostsouza, line 84: In my opinion it is still valid to see if our function indeed throws an error as expected

leonardostsouza (approved):
Added one comment, but won't block it

test/messageOperations/commonsStateMessageOperationsTest.js
b-garbacz: I changed this commentary as todo to fix

PR #147, opened 2025-08-07
Fix tests not running on bare environment

leonardostsouza:


src/utils/fileUtils.js
```diff
@@ -1,4 +1,4 @@
-import fs from 'fs';
```
b-garbacz, line 1: "bare-fs": "^4.1.6" we have already this package in our project. Please create an issue for this

b-garbacz (approved):
Do all the tests pass?

PR #174, opened 2025-08-28
Modifying indexer checks in validator and state layers

jusufsuljic:


jusufsuljic:
WHITELIST operation seems to be "broken" by the new structural changes so I was not able to fully test that functionality alone (which is not even in requirements), but just to bring awareness to it.
The crucial tests were performed manually (cli, validator and autobase layers):
Screenshot 2025-08-28 at 11 32 15
Screenshot 2025-08-28 at 11 31 33

b-garbacz (approved):
LGTM

PR #177, opened 2025-08-29
170 link writerkey

leonardotc:


src/core/state/State.js
```diff
@@ -270,6 +275,7 @@ class State extends ReadyResource {
         const initializedNodeEntry = nodeEntryUtils.init(op.cao.iw, nodeRoleUtils.NodeRole.INDEXER);
         //const updatedNodeEntry = nodeEntryUtils.setRole(initializedNodeEntry, nodeRoleUtils.NodeRole.INDEXER);
```
b-garbacz, line 276: The implementation is close to what we discussed.
We should implement a check to verify whether the `WRITER_ADDRESS` entry contains an address.

Right now, we have two cases:  
- If `WRITER_ADDRESS` is empty, then the node requesting to become a validator with the declared `op.cao.iw` can become a validator. This key can be used, and we assign it to `op.address`.  
- If `WRITER_ADDRESS` already has an assigned address, and that address is different from `op.address`, then we reject the operation.

Question for you @leonardostsouza:
If `WRITER_ADDRESS` already has an assigned address, and it’s the same as `op.address`, should we also reject the operation?

Or should we allow the following case:
If `WRITER_ADDRESS` already has an assigned address, can it reuse it?

I’m asking because someone could:  
- perform a `remove_writer` operation,  
- then an `add_writer` operation,  
- set a new writingKey,  
- set another writingKey in a different console from the same address,  
- perform another `remove_writer` in the first console from the same address,  
- and then reuse the original writingKey.

Technically it’s possible, but I’m not sure if it could lead to unexpected behavior.  
Do we want to allow users to reuse a `writingKey`?  
Or if `WRITER_ADDRESS` already has an assigned address, should we always reject the operation? 

@leonardotc join to us with your opinion, please

src/core/state/State.js
```diff
@@ -270,6 +275,7 @@ class State extends ReadyResource {
...
         await batch.put(adminAddressString, initializedNodeEntry);
+        await batch.put(EntryType.WRITER_ADDRESS + op.cao.iw.toString('hex'), adminAddressString);
```
b-garbacz, line 278: `WRITER_ADDRESS` should be assigned an address in binary format. Please use `op.address` or the `bufferToAddress` function.

src/core/state/State.js
```diff
@@ -546,6 +552,7 @@ class State extends ReadyResource {
 
         // Add the writer role to the base and update the batch
         await batch.put(requesterAddressString, updatedNodeEntry);
+        await batch.put(EntryType.WRITER_ADDRESS + op.rao.iw.toString('hex'), requesterAddressString);
```
b-garbacz, line 555: `WRITER_ADDRESS` should be assigned an address in binary format. Please use `op.address` or the `bufferToAddress` function.

b-garbacz, line 555: Missing check that I described before

src/messages/partialStateMessages/PartialStateMessageBuilder.js
```diff
@@ -170,23 +171,23 @@ class PartialStateMessageBuilder extends StateBuilder {
...
+        if (isBootstrapDeployment(this.#operationType)) {
             this.#payload.bdo = {
                 tx: tx.toString('hex'),
                 txv: this.#txValidity,
                 bs: this.#externalBootstrap,
                 in: nonce.toString('hex'),
                 is: signature.toString('hex')
             };
-        } else if (this.#isRoleAccessOperation(this.#operationType)) {
+        } else if (isRoleAccess(this.#operationType)) {
```
b-garbacz, line 182: I like it, if you did it in Partial builder, would you like to implement it complete bulder :)?

src/core/state/State.js
```diff
@@ -199,7 +200,11 @@ class State extends ReadyResource {
 
             const handler = this.#getApplyOperationHandler(op.type);
             if (handler) {
-                await handler(op, view, base, node, batch);
+                if(await this.#validateWritingKey(batch, op)) {
```
b-garbacz, line 203: Please, restore the previous logic
The handler, depending on the operation type, will select the method that should be used. This way a double check is performed unnecessarily. Additionally, with each operation you are performing a state read, which will slow down network performance under heavy load.  

Note that our methods handling operations are split into two layers:  
- the first layer performs checks that do not require a state read,  
- the second part is executed at the end and performs state reads.  

I really appreciate your code:

```
    async #validateWritingKey(batch, op) {
        const shouldValidateWritingKey = isCoreAdmin(op.type) || isRoleAccess(op.type) || isTransaction(op.type)
        if (!shouldValidateWritingKey) return true
        const address = addressUtils.bufferToAddress(op.address);
        const jsonNode = operationToPayload(op.type)
        const writingKey = b4a.toString(op[jsonNode].iw, 'hex');
        const entry = await batch.get(EntryType.WRITER + writingKey);
        return !entry || entry === address // either has no registered key or it matches the address
    }
}
```
But please keep the current structural style, because in the future we will do a light refactor. For now, I believe this is not the right approach.

b-garbacz (commented):
I’m not sure if I already mentioned this, but in `80-investigate-and-implement-better-data-packaging-of-the-transaction` we have a separated `admin_recovery` operation. This one should also include a check to verify whether the `writingKey` has already been used.

I also please, could you include this in `remove_writer`? Only the allowed address with the assigned `writingKey` should be able to remove it. I think the additional check will not be too painful. As I remember there is TODO, or something.

PR #182, opened 2025-09-01
Remove old indexer logic

jusufsuljic:


src/core/state/State.js
```diff
@@ -355,8 +352,9 @@ class State extends ReadyResource {
         // Check if the admin and indexers entry is valid
         const indexersEntry = await this.#getEntryApply(EntryType.INDEXERS, batch);
```
b-garbacz, line 353: indexer entry does not exists anymore, because we are not initializing it because of task requirements. It means that this is dead code

src/core/state/State.js
```diff
@@ -694,18 +695,15 @@ class State extends ReadyResource {
...
```
b-garbacz, line 695: indexer entry does not exists anymore, because we are not initializing it because of task requirements. It means that this is dead code

src/core/state/State.js
```diff
@@ -775,12 +773,8 @@ class State extends ReadyResource {
...
```
b-garbacz, line 773: indexer entry does not exists anymore, because we are not initializing it because of task requirements. It means that this is dead code

src/core/state/State.js
```diff
@@ -775,12 +773,8 @@ class State extends ReadyResource {
...
         if (null === indexersEntry) return;
 
-        const indexerListHasAddress = indexerEntryUtils.getIndex(indexersEntry, op.aco.ia);
-        if (indexerListHasAddress === -1) return;
-
-        // remove indexer from indexers entry
-        const updatedIndexerEntry = indexerEntryUtils.remove(op.aco.ia, indexersEntry);
-        if (updatedIndexerEntry.length === 0) return;
+        const indexerListHasWk = this.#isWriterKeyInIndexerListApply(decodedNodeEntry.wk, base);
```
b-garbacz, line 776: Btw, in all places where you are using `isWriterKeyInIndexerListApply`, you forgot to add the `await` keyword. Without `await`, you are comparing a Promise instead of the boolean result.

src/core/state/State.js
b-garbacz: all of indexersEntry are dead :p

PR #183, opened 2025-09-01
fix: Add new writer fix after store loss

b-garbacz:
Fix regression:  
If a user was a writer and lost access to the store, they should be able to remove the status of the old writer and then register a new writing key.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -758,7 +758,7 @@ export class MainSettlementBus extends ReadyResource {
                     const bootstrap_to_deploy = splitted[1];
                     await this.#handleBootstrapDeploymentOperation(bootstrap_to_deploy);
                 }
-                else if (input.startsWith("/get_wk_address")) {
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -760,7 +757,15 @@ export class MainSettlementBus extends ReadyResource {
                     const splitted = input.split(" ");
                     const bootstrap_to_deploy = splitted[1];
                     await this.#handleBootstrapDeploymentOperation(bootstrap_to_deploy);
-                } else if (input.startsWith("/get_deployment")) {
+                }
+                else if (input.startsWith("/get_wk_address")) {
```
leonardostsouza, line 761: Maybe this should be renamed to `get_validator_addr` or something like this?

leonardostsouza (approved):
Added a suggestion but won't block it

src/index.js
b-garbacz: Ok!

PR #198, opened 2025-09-05
Change nodeutils function

leonardotc:


src/utils/constants.js
```diff
@@ -36,6 +36,10 @@ export const EventType = Object.freeze({
     WARNING: 'warning',
 });
 
+// Token
+export const TOKEN_MAX = 17_000_000n
```
leonardostsouza, line 40: In theory, we won't have a max token count. We estimate there are currently 17kk, but this is still an approximation, not a hard limit

This shouldn't be set as a constant.

test/state/stateTestUtils.js
```diff
@@ -9,4 +10,13 @@ export function randomBuffer(size) {
 export function randomAddress(hrp = TRAC_NETWORK_MSB_MAINNET_PREFIX) {
     const data = randomBuffer(TRAC_PUB_KEY_SIZE);
     return bech32m.encode(hrp, bech32m.toWords(data));
-}
\ No newline at end of file
+}
+
+export const ZERO_BALANCE = b4a.alloc(BALANCE_BYTE_LENGTH)
+
+export const TEN_THOUSAND = Buffer.from([
```
leonardostsouza, line 17: `Buffer.from` won't work on Bare. Please, use `b4a.from`

leonardostsouza: Also, the balance is a binary representation of integer part + decimal part. When someone has `10000` $TNK, the balance should actually be `10000` followed by 18 more zeroes: `10_000_000_000_000_000_000_000`. In hex, this would be `0x21E19E0C9BAB2400000`

leonardotc: 10 thousand there aint necessarily 10 thousand tokens

src/core/state/utils/nodeEntry.js
```diff
@@ -53,9 +53,10 @@ export function encode(node) {
     }
 
     try {
-        const entry = b4a.alloc(1 + node.wk.length);
+        let entry = b4a.alloc(NODE_ENTRY_SIZE);
         entry[0] = nodeRole;
-        b4a.copy(node.wk, entry, 1);
+        entry = setWritingKey(entry, node.wk);
```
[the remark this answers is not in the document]
leonardotc: Yes then the buffer wont be valid and that will return null as well.

leonardostsouza (approved):

PR #199, opened 2025-09-05
186 story balance operations

leonardotc:


leonardotc:
This requires the other 186 to be merged (which is why there is a conflict)

leonardostsouza (approved):
LGTM

b-garbacz:
is it ready for review?

src/core/state/utils/balance.js
```diff
@@ -0,0 +1,90 @@
...
+
+    update(nodeEntry) {
+        return setBalance(nodeEntry, this.#value)
+    }
+
+    equals(b) {
+        return b4a.equals(this.#value, b.value)
+    }
+
+    greaterThen(b) {
```
b-garbacz, line 75: Typo: greatedThan or better - `gt`

src/core/state/utils/balance.js
```diff
@@ -0,0 +1,90 @@
...
+        return b4a.compare(this.#value, b.value) === 1
+    }
+
+    lowerThen(b) {
```
b-garbacz, line 79: Typo: lower than or better - `lt`

src/core/state/utils/balance.js
```diff
@@ -0,0 +1,90 @@
...
+        return b4a.compare(this.#value, b.value) === -1
+    }
+
```
b-garbacz, line 82: We can add `lte` and `gte`

src/core/state/utils/balance.js
```diff
@@ -0,0 +1,90 @@
...
+const addBuffers = (a, b) => {
+    if (a.length !== b.length) return EMPTY_BUFFER
+    const result = b4a.alloc(a.length);
+    let carry = 0;
+    for (let i = a.length - 1; i >= 0; i--) {
+        const sum = a[i] + b[i] + carry;
+        result[i] = sum & 0xff;
+        carry = sum >> 8;
+    }
+    return result;
```
b-garbacz, line 22: @leonardostsouza  What if there will be an overflow? Maybe we should return empty buffer if carry is >0?

src/core/state/utils/balance.js
```diff
@@ -0,0 +1,90 @@
...
+        let diff = a[i] - b[i] - borrow;
+        if (diff < 0) {
+            diff += 256;
+            borrow = 1;
+        } else {
+            borrow = 0;
+        }
+        result[i] = diff;
+    }
+    return result;
```
b-garbacz, line 39: If borrow is still 1 after processing all bytes, it indicates an underflow. We probably don`t want negative results right? Maybe we should return empty buffer if borrow is > 0?

src/core/state/utils/balance.js
```diff
@@ -0,0 +1,90 @@
...
+}
+
+const validateValue = value => {
+    if (!isBufferValid(value, BALANCE_BYTE_LENGTH)) {
+        throw new BalanceError(value)
```
b-garbacz, line 44: It will be used in apply function. I will write a reminder: Apply function must be throw resistant...

b-garbacz (changes requested):

src/core/state/utils/balance.js
leonardostsouza: I actually prefer the bigger name. It's intuitive when you read it in code.

src/core/state/utils/balance.js
jusufsuljic: longer name left.

src/core/state/utils/balance.js
leonardostsouza: Maybe we can create an an alias that will invoke `lowerThan` or `greaterThan` but, in any case, it would be better to use the bigger names for the sake of code readability.

src/core/state/State.js
```diff
@@ -120,6 +121,25 @@ class State extends ReadyResource {
         return nodeEntry ? nodeEntryUtils.decode(nodeEntry) : null;
     }
 
+    async incrementBalance(address, toIncrement) {
+        if (isBufferValid(toIncrement, NODE_ENTRY_SIZE) || b4a.equals(toIncrement, ZERO_BALANCE)) return null
+        const nodeEntry = await this.getNodeEntry(address);
```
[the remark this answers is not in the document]
jusufsuljic: well the question is: do we ever need to directly modify the balance from the state? @leonardostsouza @b-garbacz  Asking because we don't want to simply recreate these 2 functions for apply fn if they will not be used at all.

test/state/utils/balance.test.js
```diff
@@ -0,0 +1,101 @@
+import { test } from 'brittle';
+import b4a from 'b4a';
```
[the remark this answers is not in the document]
jusufsuljic: You can check the latest for new tests.

src/core/state/State.js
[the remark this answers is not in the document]
b-garbacz: I understand your concerns, but when reading the contract state, we need to use a batch from the linearizer to maintain determinism, as shown below:

```js
async #getEntryApply(key, batch) {
    const entry = await batch.get(key);
    return entry !== null ? entry.value : null;
}

For off-chain reads, you can use the methods described above the disclaimer. However, for apply, we need to implement everything separately. We did something like this before with Leo (we were using get and getSigned directly), and Benny ended up fixing the entire code for us.

src/core/state/utils/balance.js
jusufsuljic: I addressed this, can you please review the latest?

src/core/state/utils/balance.js
b-garbacz: @jusufsuljic alright but I wanted to suggest to add less than operation or equal and greater than or equal operation because they might be useful

jusufsuljic: well we can add those, yes.

leonardostsouza (approved):

PR #201, opened 2025-09-08
rpc server running on msb node via https & bare

lucasfernandes:
**What is done so far**
Integrating the RPC server directly into the main node (msb.mjs). This allows both the RPC and the MSB to run in a single process, giving the RPC access to the database.

To make it portable and secure, I’m now using the standard https module from Node.js, which is made to work in the Pear environment through its compatibility packages. This keeps the code clean and portable across different runtimes.
 The RPC server is now secured with SSL and can be toggled on or off with a command-line argument. We've also set up a safe way to handle local development certificates.  

**How to**  

Certificates: To generate certificates to run locally: 
openssl req -nodes -new -x509 -keyout key.pem -out cert.pem

To start MSB with RPC - the main one: 
npm run dev-rpc —store=  

To start MSB without RPC remains the same: 
npm run dev

To call the endpoint: 
curl -k https://localhost:3000/balance/

**Important:**

The (-k) flag is used because you are connecting to an HTTPS server that is using a self-signed SSL certificate. Without this flag, curl (and web browsers) will not trust the connection and will return an error. It will not be necessary when we have the official one.


package.json
```diff
@@ -1,52 +1,54 @@
...
+    },
+    "devDependencies": {
+        "bare-os": "^3.6.1",
+        "bip39-mnemonic": "^2.4.0",
+        "protocol-buffers": "^4.2.0"
+    },
+    "publishConfig": {
+        "registry": "https://registry.npmjs.org",
+        "access": "public"
+    }
```
leonardostsouza, line 53: Git has detected too many changes in `package.json`, while only a few lines are really different. Did indentation change?

rpc_server.mjs
```diff
@@ -0,0 +1,45 @@
...
+// SSL Certifications
+const sslOptions = {
+    key: fs.readFileSync('./key.pem'),
+    cert: fs.readFileSync('./cert.pem'),
+}
+
+// Called by msb.mjs file
+export function startRpcServer(msbInstance) {
+    const server = https.createServer(sslOptions, async (req, res) => {
+        if (req.url.startsWith('/balance/')) {
```
leonardostsouza, line 14: It would be good to export these handlers to a separate file, otherwise this one will become very bloated, very quickly. But for now, this is ok

src/index.js
```diff
@@ -1,395 +1,347 @@
 /** @typedef {import('pear-interface')} */ /* global Pear */
-import ReadyResource from "ready-resource";
```
leonardostsouza, line 2: I believe this whole file was parsed by some local linter and it changed a lot of lines. I agree with most of them, but it would be better if they were n a separate PR

leonardostsouza (approved):
LGTM

PR #202, opened 2025-09-08
200 implement disable initialization and balance initialization   offchain part

b-garbacz:
Interface of Admin:
Screenshot 2025-09-08 at 19 22 15
Interface of regular user:
Screenshot 2025-09-08 at 19 23 03


src/index.js
```diff
@@ -44,13 +44,15 @@ export class MainSettlementBus extends ReadyResource {
     #enable_auto_transaction_consent
     #state;
     #isClosing = false;
+    #is_admin_node;
```
leonardostsouza, line 47: Just a matter of style, but I would call it `is_admin_mode` to make it clear that this is just the way you initialized the process, and has nothing to do with how the network checks if a node is indeed an admin or not

src/index.js
```diff
@@ -44,13 +44,15 @@ export class MainSettlementBus extends ReadyResource {
...
 
     constructor(options = {}) {
         super();
         this.#stores_directory = options.stores_directory;
         this.#key_pair_path = `${this.#stores_directory}${options.store_name}/db/keypair.json`;
         this.#enable_wallet = options.enable_wallet !== false;
         this.enable_interactive_mode = options.enable_interactive_mode !== false;
+        this.#is_admin_node = options.store_name === 'admin';
```
leonardostsouza, line 55: Read commentary above

leonardostsouza (approved):

src/index.js
b-garbacz: Oki I will change it

PR #231, opened 2025-09-16
Adding burn function to Balance

jusufsuljic:


leonardotc:
This will be solved by #236

PR #234, opened 2025-09-16
227 implement fee handling logic for bootstrap deployment tx add admin remove indexer and ban validator operations

b-garbacz:


src/core/state/utils/balance.js
```diff
@@ -222,6 +230,17 @@ class Balance {
         return toBalance(result)
     }
 
+    /**
+     * Reduce to its percentage
+     * @param {Buffer} percent - the buffer percentage (with two extra decimals). Expected to be used along with percent function.
+     * @returns {Balance} - New Balance instance
+     */
+    percentage(percent) {
```
jusufsuljic, line 238: This one is double, the one above is the correct one. Just remove this one.

jusufsuljic (changes requested):

PR #236, opened 2025-09-17
Fees

leonardotc:


src/core/state/State.js
```diff
@@ -529,6 +529,27 @@ class State extends ReadyResource {
...
+        if (validatorNodeEntry === null) return;
+        const validatorBalance = toBalance(validatorNodeEntry.balance);
+        if (validatorBalance === null) return;
+        const newValidatorBalance = validatorBalance.add(feeAmount.percentage(FEE_REWARD));
+        if (newValidatorBalance === null) return;
+        const updatedValidatorNodeEntry = newValidatorBalance.update(validatorNodeEntry)
+        if (updatedValidatorNodeEntry === null) return;
+
         const isNewWkInIndexerList = await this.#isWriterKeyInIndexerListApply(op.rao.iw, base);
         if (isNewWkInIndexerList) return; // New admin wk is already in indexers entry
```
b-garbacz, line 554: Can you move this check before portioning fee? 2 lines above.

src/core/state/State.js
```diff
@@ -529,6 +529,27 @@ class State extends ReadyResource {
         const adminNodeEntry = await this.#getEntryApply(requesterAdminAddressString, batch);
         const newAdminNodeEntry = setWritingKey(adminNodeEntry, op.rao.iw)
 
+        // Fee
```
b-garbacz, line 532: Hey questions
- Where is balance check `if (!rgreaterThanOrEquals(feeAmount)) return;`. What if the user does not have sufficient founds?
- did you test admin recovery manually and ensured that?

I paused the review here, because if the requester’s balance is not validated in this part, it’s likely missing in the following handlers as well.

src/core/state/State.js
```diff
@@ -529,6 +529,27 @@ class State extends ReadyResource {
...
+        const chargedAdminEntry = updatedFee.update(newAdminNodeEntry)
+
+        // Reward logic
+        const validatorNodeEntryBuffer = await this.#getEntryApply(validatorAddressString, batch);
+        if (validatorNodeEntryBuffer === null) return;
+        const validatorNodeEntry = nodeEntryUtils.decode(validatorNodeEntryBuffer);
+        if (validatorNodeEntry === null) return;
+        const validatorBalance = toBalance(validatorNodeEntry.balance);
+        if (validatorBalance === null) return;
+        const newValidatorBalance = validatorBalance.add(feeAmount.percentage(FEE_REWARD));
```
b-garbacz, line 548: This constant is a little bit confusing because for this operation this 75% but for another operations i can be 50%

src/core/state/utils/balance.js
```diff
@@ -38,15 +72,7 @@ const addBuffers = (a, b) => {
     return result;
 }
 
-/**
- * Subtracts buffer b from buffer a as unsigned integers.
- * Returns a new buffer containing the result.
- * Underflow wraps around modulo 2^(length*8).
- * @param {Buffer} a
```
b-garbacz, line 45: why did you return docs?

src/core/state/utils/balance.js
```diff
@@ -17,13 +34,30 @@ export const $TNK = bigint => bigIntToBuffer(
...
+ * @returns {Buffer} - Fixed-length buffer representing token amount
  */
+export const toTerm = bigint => bigIntToBuffer(
+    bigint, 
+    BALANCE_BYTE_LENGTH
+)
+
+const truncate = buf => buf.slice(BALANCE_BYTE_LENGTH * -1)
+
+// Thank you gpt
```
b-garbacz, line 49: I don't know how is responsible for this commentary. However. This part of the code seems to be a critical operation (bitwuse shift on balances in apply function - which is by default critical for us). Adding informal comments such as "thank you gpt" may give the wrong impression to external reviewers, auditors, or even potential partners who read the code. 

I don't know if this is how we show that we are professional and trustworthy, which is very important to us. I recommend either removing this comment or replacing it with A clear technical explanation like for another functions. . This will make the codebase easier to maintain and also reflect better on the overall quality of the project.

src/core/state/utils/balance.js
```diff
@@ -69,6 +95,60 @@ export const subBuffers = (a, b) => {
     return result;
 }
 
+const divBuffers = (dividend, divisor) => {
+    if (dividend.length !== BALANCE_BYTE_LENGTH || divisor.length !== BALANCE_BYTE_LENGTH) {
```
b-garbacz, line 99: missing docs

b-garbacz (changes requested):

src/core/state/utils/balance.js
leonardostsouza: Ok. I have 2 questions:

1. Why is it private now?
2. Even if this is now a private function, if the docs were already there, wouldn't it be a good idea to keep them? The comments may also help other developers maintain the code in the future.

src/core/state/utils/balance.js
leonardostsouza: Yeah, I agree. Specially because this is open-source code

src/core/state/utils/balance.js
```diff
@@ -69,6 +95,60 @@ export const subBuffers = (a, b) => {
...
+        const ai = a[i];
+        const bj = b[j];
+
+        const pos = i + j + 1;
+        const mul = ai * bj + result[pos] + carry;
+
+        result[pos] = mul & 0xff;   // low byte
+        carry = mul >>> 8;          // high byte
+        }
+        result[i] += carry;
```
b-garbacz, line 145: @leonardostsouza won't that cause overflow?

src/core/state/utils/balance.js
```diff
@@ -127,6 +198,36 @@ class Balance {
...
+        return toBalance(result)
+    }
+
+    /**
+     * Reduce to its percentage
+     * @param {Buffer} percent - the buffer percentage (with two extra decimals). Expected to be used along with percent function.
+     * @returns {Balance} - New Balance instance
+     */
+    percentage(percent) {
+        const dividend = mulBuffers(this.#value, percent)
```
b-garbacz, line 217: Don't we need input validation?

src/core/state/State.js
leonardotc: The code is kind of safe against that because the subtraction would return the `NULL_BUFFER` which would in turn result in a null balance and subsequently aborted. This is one of the scenarios that we tested during the `APPEND_WHITELIST` but we can add the check, yes.

src/core/state/utils/balance.js
b-garbacz: Thanks for adding the test. Just to clarify , tests should be written for the project itself for example to ensure long term stability and prevent regressions, not specifically for me or any review. Future contributors can rely on them and our team with confidence. 

I will keep this discussion unresolved for now, as I would also like to hear @leonardostsouza’s opinion to confirm everything is correct on his side. For calculations, I prefer to perform double checks.

src/core/state/utils/balance.js
jusufsuljic: I think in order to be consistent and also follow some standards, we should document the exported public members only. We removed comments from private members because of this,

leonardostsouza (approved):

b-garbacz (approved):

PR #245, opened 2025-09-18
238 configure cors on rpc server

lucasfernandes:
**Pull Request Overview**

This pull request breaks down the existing `rpc_server.mjs` file into a more modular and maintainable structure. The goal is to improve code organization, making it easier to manage and scale the API in the future.

**Summary of Changes**
This change refactors the RPC server from a single, monolithic file into a new, organized directory with clear separation of concerns. The core logic is now distributed across four files:

**cors.mjs:** Contains the logic for handling `Cross-Origin Resource Sharing (CORS)` headers and preflight OPTIONS requests. This keeps security concerns isolated from the main application logic.

**handlers.mjs:** Houses all the handler functions. Each function is responsible for the business logic of a single API endpoint (e.g., fetching a balance, broadcasting a transaction). This makes it easy to add or modify endpoints without changing the core server file.

**router.mjs:** Serves as the central "map" for the server. It defines all the available routes by mapping an HTTP method and path to a specific handler function from `handlers.mjs`.

**server_rpc.mjs:** The main entry point for the server. It instantiates the HTTPS server, applies the CORS logic, and then uses the `router.mjs` file to dynamically route incoming requests to the correct handler.

This new structure is more scalable and adheres to best practices for building robust Node.js applications.

### DIFF
--- a/rpc/cors.mjs
+++ b/rpc/cors.mjs
@@ -1,0 +1,15 @@
+export function applyCors(req, res) {
+    // Set CORS headers
+    res.setHeader('Access-Control-Allow-Origin', '*'); // Allow all origins, or specify a domain
+    res.setHeader('Access-Control-Allow-Methods', 'GET, POST, OPTIONS');
+    res.setHeader('Access-Control-Allow-Headers', 'Content-Type');
+
+    // Handle preflight OPTIONS request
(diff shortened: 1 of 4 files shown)


leonardostsouza (approved):
LGTM

PR #246, opened 2025-09-18
https to http

lucasfernandes:
**Pull Request Overview**

This pr updates the rpc server to use **http** instead of https.
Now the certificates are not required anymore and we should implement a reverse proxy (another task)


b-garbacz (approved):
For me it looks good

jusufsuljic:
Look good, but one question: is it better to detect bare environment and then enforce bare specific ones? It seems like for local testing and development we always need to comment out and replace these packages because they throw because we run the app mostly with Node locally.

Just an idea. @lucasfernandes @b-garbacz @leonardostsouza

leonardostsouza (approved):

leonardostsouza (changes requested):

PR #251, opened 2025-09-19
Exposing endpoint for fetching unconfirmed balance by address

jusufsuljic:


jusufsuljic:
@lucasfernandes @b-garbacz @leonardostsouza Note: We could also merge these 2 handlers in 1 for confirmed/unconfirmed balance and control the output with this flag. It is just an optimization tor reduce the amount of code that serves a similar/same purpose.

PR #263, opened 2025-09-23
Applied safe logging to all op handlers

jusufsuljic:


src/core/state/State.js
```diff
@@ -303,115 +303,208 @@ class State extends ReadyResource {
...
+            this.#enable_txlogs && this.#safeLogApply(OperationType.BALANCE_INITIALIZATION, "Balance initialization is disabled.", node.from.key)
+            return;
+        };
 
         // Ensure that an admin invoked this operation
         const adminEntry = await this.#getEntryApply(EntryType.ADMIN, batch);
         const decodedAdminEntry = adminEntryUtils.decode(adminEntry);
-        if (null === decodedAdminEntry || !this.#isAdminApply(decodedAdminEntry, node)) return;
+
+        // NOTE: Would it make sense to extract null === decodedAdminEntry and log the error?
```
leonardostsouza, line 358: What do you mean exactly by "extract null === decodedAdminEntry"?

jusufsuljic: Like to separate the errors where: the decodedAdminNetry is null - meaning the decoding failed, and below just do an another if with the other !this.#isAdminApply.... and there log tht a non admin tried to do the operation. Again, just pure cometics to point users to the issue faster.

leonardostsouza: It makes sense, but it is not necessary, imo

leonardostsouza (approved):
LGTM

src/core/state/State.js
jusufsuljic: okay cool.

b-garbacz:
LGTM

PR #265, opened 2025-09-23
fix state mutation issue

b-garbacz:
## Status
Critical 

## Problem
Unexpected state mutations caused by shared buffer references returned from `batch.get()`

## Solution
- Added deep buffer copying in `#getEntryApply`,  `#getDeploymentEntryApply` and `#getRegisteredWriterKeyApply` methods

## Technical Details
State now changes only through explicit `batch.put()` operations, preventing accidental mutations of shared buffer references.
The issue was that multiple parts of the code were modifying the same buffer instances in memory. By implementing deep copying at the data retrieval level, each operation now works with isolated copies, maintaining data integrity.

### DIFF
--- a/src/core/state/State.js
+++ b/src/core/state/State.js
@@ -1804,12 +1804,12 @@ class State extends ReadyResource {
 
     async #getEntryApply(key, batch) {
         const entry = await batch.get(key);
-        return entry !== null ? deepCopyBuffer(entry.value) : null
+        return deepCopyBuffer(entry?.value)
     }
     
     async #getDeploymentEntryApply(key, batch) {
         const entry = await batch.get(EntryType.DEPLOYMENT + key);
-        return entry !== null ? deepCopyBuffer(entry.value) : null
+        return deepCopyBuffer(entry?.value)
     }
 
     async #getIndexerSequenceStateApply(base) {
@@ -1836,7 +1836,7 @@ class State extends ReadyResource {
 
     async #getRegisteredWriterKeyApply(batch, writingKey) {
         const entry =  await batch.get(EntryType.WRITER_ADDRESS + writingKey);
-        return entry !== null ? deepCopyBuffer(entry.value) : null
+        return deepCopyBuffer(entry?.value)
     }
 
     async #isApplyInitalizationDisabled(batch) {
--- a/src/core/state/State.js
+++ b/src/core/state/State.js
@@ -23,6 +23,7 @@ import { blake3Hash } from '../../utils/crypto.js';
 import { BALANCE_FEE, toBalance, PERCENT_75, PERCENT_50, PERCENT_25 } from './utils/balance.js';
 import { safeWriteUInt32BE } from '../../utils/buffer.js';
 import deploymentEntryUtils from './utils/deploymentEntry.js';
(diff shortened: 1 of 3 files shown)


jusufsuljic (approved):

src/utils/buffer.js
```diff
@@ -6,7 +6,7 @@ export const NULL_BUFFER = b4a.alloc(0) // null buffer (single byte of 0)
 
 const isUInt32 = (n) => { return Number.isInteger(n) && n >= 1 && n <= 0xFFFFFFFF; }
 
-export const isNullBuffer = (buffer) => {b4a.equals(buffer, NULL_BUFFER)}
+export const isNullBuffer = (buffer) => { b4a.equals(buffer, NULL_BUFFER) }
```
leonardotc, line 9: is this correct? Cause i think its missing a return statement which is low-key dangerous.

src/utils/buffer.js
```diff
@@ -42,16 +42,23 @@ export function normalizeBuffer(message) {
...
+
     return null;
 }
 
-export const bigIntToBuffer = bigIntTo16ByteBuffer
\ No newline at end of file
+export const bigIntToBuffer = bigIntTo16ByteBuffer
+
+export function deepCopyBuffer(buffer) {
+    if (!buffer) return null;
```
leonardotc, line 60: `NULL_BUFFER`?

src/utils/buffer.js
b-garbacz: Didn't you write this part?

src/core/state/State.js
```diff
@@ -1752,7 +1753,6 @@ class State extends ReadyResource {
         if (!isSelfTransfer) {
             const recipientEntryBuffer = await this.#getEntryApply(recipientAddressString, batch);
             if (recipientEntryBuffer === null) {
-                if (transferAmount.value === null) return null;
```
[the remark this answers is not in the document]
b-garbacz: I am not sure if I did this typo on last Thursday, it was 1 AM

src/utils/buffer.js
b-garbacz: I think it can be risky. Please notice that in many places in the code when we want to fetch entry thought `result = getEntryApply(...)` which uses  deepCopyBuffer(), we perorm security check `if(restul === null) {skip}`. If I will change null to `NULL_BUFFER ` I will need to go over the whole code  in `state.js` for this specific situation. Can we stay with null?

leonardotc: Ofc.

src/core/state/State.js
```diff
@@ -1835,7 +1835,8 @@ class State extends ReadyResource {
     }
 
     async #getRegisteredWriterKeyApply(batch, writingKey) {
-        return await batch.get(EntryType.WRITER_ADDRESS + writingKey);
+        const entry =  await batch.get(EntryType.WRITER_ADDRESS + writingKey);
+        return entry !== null ? deepCopyBuffer(entry.value) : null
```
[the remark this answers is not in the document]
leonardotc: Sure. is all good.

b-garbacz:
@leonardotc @jusufsuljic I have added unit tests for deep copy function

PR #290, opened 2025-10-07
Code refactoring and logging added

jusufsuljic:


jusufsuljic:
Screenshot 2025-10-07 at 15 46 47
Screenshot 2025-10-07 at 15 46 34

jusufsuljic:
In essence, we did not change anything critical, the behavior in case this error happens remains the same - the batch.put updates will not trigger and the handler will proceed.
However now we have this unavoidable system log in case this ever happens (should **NEVER** happen)
@b-garbacz @leonardostsouza

PR #291, opened 2025-10-08
Improve validator consistency check

b-garbacz:


src/core/state/State.js
```diff
@@ -1296,8 +1297,14 @@ class State extends ReadyResource {
             return Status.FAILURE;
         };
 
+        const validatorEntryBuffer = await this.#getEntryApply(validatorAddressString, batch);
+        if (validatorEntryBuffer === null) {
```
leonardostsouza, line 1301: We are repeating this check in many places. Can't we put this inside the function `#isValidatorValidApply`?

src/core/state/State.js
```diff
@@ -1578,12 +1580,18 @@ class State extends ReadyResource {
...
```
leonardostsouza, line 1584: See comment above

src/core/state/State.js
```diff
@@ -2440,8 +2442,14 @@ class State extends ReadyResource {
             return Status.FAILURE;
         };
 
+        const validatorNodeEntryBuffer = await this.#getEntryApply(validatorAddressString, batch);
+        if (validatorNodeEntryBuffer === null) {
```
leonardostsouza, line 2446: Same as above

src/core/state/State.js
```diff
@@ -2671,8 +2673,14 @@ class State extends ReadyResource {
             return Status.FAILURE;
         };
 
+        const validatorEntryBuffer = await this.#getEntryApply(validatorAddressString, batch);
+        if (validatorEntryBuffer === null) {
```
leonardostsouza, line 2677: Same as above

src/core/state/State.js
```diff
@@ -2865,8 +2874,14 @@ class State extends ReadyResource {
...
```
leonardostsouza, line 2878: Same as above

src/core/state/State.js
b-garbacz: Done

PR #320, opened 2025-10-18
Bug: Can not sync with network when wallet is disabled

b-garbacz:
When MSB was started in reader mode, the node failed to synchronize with the network because the PoolService required a wallet address to operate. This caused the synchronization process to stop entirely, preventing the node from joining the network in reader mode.

## What was done
- Added a wallet flag check before starting the `PoolService`.  
- If the wallet is inactive, the service does not start. Otherwise, it runs normally.  
- Refactored the class to use a non-blocking Scheduler instead of a blocking loop or setInterval.
- Refactored the class into smaller methods and improved readability.
- Removed unused files related to the old implementation.

### DIFF
--- a/src/core/network/services/PoolService.js
+++ b/src/core/network/services/PoolService.js
@@ -59,7 +59,7 @@ class PoolService {
     }
 
     #createScheduler() {
-        return new Scheduler(() => this.#worker(), PROCESS_INTERVAL_MS);
+        return new Scheduler((next) => this.#worker(next), PROCESS_INTERVAL_MS);
     }
 
     async #processTransactions() {


src/utils/Scheduler.js
```diff
@@ -0,0 +1,94 @@
...
+    #isRunning = false;
+    #timer = null;
+    #currentWorkerRun = null;
+
+    constructor(worker, defaultInterval) {
+        if (typeof worker !== 'function') {
+            throw new TypeError('Worker must be a function');
+        }
+        this.#worker = worker;
+        this.#defaultInterval = defaultInterval ?? 100;
```
leonardostsouza: Should `0` be considered a valid interval?

src/utils/Scheduler.js
```diff
@@ -0,0 +1,94 @@
+class Scheduler {
```
leonardotc: Add some docs. I dont think we need a very extensive js docs but there are three entities of time here: one that temporizes an initial delay and the others that are associated with schedule. I can sort of guess as to why: you may want to set a delay when you start to initialize resources, you want a default interval that will be the one that is actually used and you may want to change that in run time to implement back-off-like rules. Wiring a comment about the reasoning is probably a good idea.

src/utils/Scheduler.js
```diff
@@ -0,0 +1,94 @@
...
+    async run() {
+        if (!this.isRunning) return null;
+
+        let scheduleCalled = false;
+        let nextDelay = null;
+
+        const scheduleNext = (ms) => {
+            const delayMs = Number(ms);
+
+            if (!Number.isFinite(delayMs) || delayMs < 0) {
```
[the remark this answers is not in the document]
b-garbacz: Oki I will create a static function which will validate it and I will place in serval of places

PR #333, opened 2025-10-23
Fix fee validation in role validation - roleOperationValidator

b-garbacz:
During the system audit, I noticed that on the validator side, during the initial validation for the **add writer** operation, we were not checking whether the user had enough stake balance to cover the transaction. This caused the validator to be penalized for an invalid transaction, since the protocol did not accept it.

I implemented a method that verifies this condition for both the signed and unsigned versions. I also added guards in the CLI interface so that the user can see what they are doing incorrectly.

Additionally, in the **/help** section, I described how much will be charged depending on the operation, to prevent situations where the user is surprised by a higher fee.

While testing the code in various scenarios, I also improved the user feedback messages in **state.js**.

### DIFF
--- a/src/utils/cli.js
+++ b/src/utils/cli.js
@@ -64,27 +64,27 @@ export function printHelp(isAdminMode = false) {
     if (isAdminMode) {
         console.log('🚨 WARNING: IF YOU ARE NOT AN ADMIN, DO NOT USE THE COMMANDS BELOW! YOU RISK LOSING YOUR FUNDS! 🚨');
         console.log('\nAdmin commands:');
-        console.log('- /add_admin: register admin entry with bootstrap key (initial setup), or use --recovery flag to recover admin role');
-        console.log('- /balance_migration: perform balance migration with the given initial balances CSV file');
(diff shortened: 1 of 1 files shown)


src/core/state/State.js
```diff
@@ -2062,7 +2082,12 @@ class State extends ReadyResource {
         };
 
         const decodedAdminEntry = adminEntryUtils.decode(adminEntry);
-        if (decodedAdminEntry === null || !this.#isAdminApply(decodedAdminEntry, node)) {
+        if (decodedAdminEntry === null) {
+            this.#safeLogApply(OperationType.REMOVE_INDEXER, "Failed to decode admin entry.", node.from.key)
+            return Status.FAILURE;
+        };
+
+        if (!this.#isAdminApply(decodedAdminEntry, node)) {
```
leonardostsouza: Could you please extract all these instances of repeated code in a "verifyAdminEntry" function? It will be easier to maintain and the code will even be safer if you do it

leonardostsouza (approved):

jusufsuljic (approved):
One question, you wrote in the description "stake balance", this has nothing to do with staked, this is regular token balance correct? Or I got it wrong?

PR #337, opened 2025-10-26
Setting a new ignore status for non-penalty scenarios

b-garbacz:


src/core/state/State.js
```diff
@@ -2589,18 +2595,20 @@ class State extends ReadyResource {
...
 
         // anti-replay attack
+        // NOTE: Also here, attempt at bootstrap deployment with the same payload is suspicious so we should punish the second attempt in the same batch. 
+        // Does this make sense?
         const hashHexString = op.bdo.tx.toString('hex');
         const opEntry = await this.#getEntryApply(hashHexString, batch);
         if (opEntry !== null) {
             this.#safeLogApply(OperationType.BOOTSTRAP_DEPLOYMENT, "Operation has already been applied.", node.from.key)
-            return Status.FAILURE;
+            return Status.IGNORE;
```
leonardostsouza: If this is suspicious, shouldn't we return FAILURE, instead?

PR #351, opened 2025-11-03
Shorting the balance when penalty amt is greater than the staked amt

jusufsuljic:


b-garbacz (approved):
LGTM. With this fix we are never risking with underflow.

PR #355, opened 2025-11-04
354 enhance migration script

b-garbacz:


src/utils/fileUtils.js
```diff
@@ -13,16 +17,36 @@ import { decimalStringToBigInt, bigIntTo16ByteBuffer, bufferToBigInt } from './a
...
+        for (const address of addresses) {
+            if (seen.has(address)) {
+                duplicates.add(address);
+            } else {
+                seen.add(address);
+            }
+        }
+
+        if (duplicates.size > 0) {
+            throw new Error(`Duplicate addresses found in whitelist file: ${Array.from(duplicates).join(', ')}. Each address must be unique.`);
```
jusufsuljic: Just an idea - if user adds 1000 addresses in the list, and he has 2 duplicates somewhere, and we throw this exception - it would probably be (if user has normal intention) tedious for him to locate these duplicates.

Having the idea to maybe do a whitelist on top of the "seen" list, and ignore the duplicates, and perform the whitelist as expected.

Maybe we can log like duplicates ignored or similar as a message, but keep the operation running.

What do you think?

src/utils/fileUtils.js
```diff
@@ -90,7 +115,113 @@ async function readBalanceMigrationFile(filepath = BALANCE_MIGRATION_FILEPATH) {
...
+    }
+}
+
+export async function validateMigrationData(addresses, migrationDirectory = BALANCE_MIGRATED_DIR) {
+    const migrationFiles = await getAllMigrationFiles(migrationDirectory);
+    if (migrationDirectory === BALANCE_MIGRATED_DIR) {
+        for (let i = 0; i < addresses.length; i++) {
+            const address = addresses[i].address;
+            for (const filePath of migrationFiles) {
+                const content = await fs.promises.readFile(filePath, 'utf8');
```
leonardotc: Its a small optimization but for each address, you will read all files. It would be better, for each file, to traverse the list in memory (since file reading is IO).

b-garbacz: Done

PR #373, opened 2025-11-13
feat:rpc-msb-dockerization

b-garbacz:


src/core/network/identity/IdentityProvider.js
```diff
@@ -0,0 +1,111 @@
+import PeerWallet from 'trac-wallet';
+import { TRAC_NETWORK_MSB_MAINNET_PREFIX } from 'trac-wallet/constants.js';
+import b4a from 'b4a';
+
+class IdentityProvider {
```
leonardotc: Ok so, provider as far as I remember is pretty much a factory, its odd to call it here when its just an object that proxies to a strategy. In this case I would call it Wrapper or something. Identity is also something that is pretty strict in our world. Its stuff that I would either associated with a domain model with dids.

src/core/network/Network.js
```diff
@@ -239,6 +243,16 @@ class Network extends ReadyResource {
             console.log(e)
         }
     }
+
+    #resolveIdentityProvider(keyPair, wallet) {
+        if (!this.#identityProvider) {
+            this.#identityProvider = this.#enable_wallet
```
leonardotc: Deciding which instance to load is the responsibility of a provider. A provider is pretty much a factory (that aint complex enough to be abstract).

package.json
```diff
@@ -9,9 +9,12 @@
   "type": "module",
   "scripts": {
     "dev": "pear run -d .",
-    "prod": "NODE_OPTIONS='--max-old-space-size=4096' pear run . ${npm_config_store}",
     "dev-rpc": "pear run -d . ${npm_config_store} --rpc --port ${npm_config_port}",
-    "prod-rpc": "pear run . ${npm_config_store} --rpc --host ${npm_config_host} --port ${npm_config_port}",
+    "prod": "NODE_OPTIONS='--max-old-space-size=4096' pear run . ${npm_config_store}",
+    "prod-rpc": "NODE_OPTIONS='--max-old-space-size=4096' pear run . ${npm_config_store} --rpc --host ${npm_config_host} --port ${npm_config_port}",
+    "env-prod": "if [ -f .env ]; then set -a; . ./.env; set +a; fi; NODE_OPTIONS='--max-old-space-size=4096' pear run. ${MSB_STORE:-node-store}",
+    "env-prod-rpc": "if [ -f .env ]; then set -a; . ./.env; set +a; fi; NODE_OPTIONS='--max-old-space-size=4096' pear run . ${MSB_STORE:-rpc-node-store} --rpc --host ${MSB_HOST:-0.0.0.0} --port ${MSB_PORT:-5000}",
```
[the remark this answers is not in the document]
b-garbacz: default port :x

leonardotc (approved):

PR #376, opened 2025-11-14
msb dockerization with not working pear runtime

b-garbacz:


src/core/network/identity/IdentityProvider.js
```diff
@@ -0,0 +1,111 @@
+import PeerWallet from 'trac-wallet';
+import { TRAC_NETWORK_MSB_MAINNET_PREFIX } from 'trac-wallet/constants.js';
+import b4a from 'b4a';
+
+class IdentityProvider {
```
leonardotc: The problem with this is that this aint exactlu a provider, it just delegates. In general, you want something to load and return a strategy to the consumer class. A provider is a... kind of... creational pattern (it is an in-between factory method and abstract factory, i guess). This is like a proxy which falls into: This is https://sourcemaking.com/refactoring/smells/middle-man strategies work cause they share the same interface and therefore dont need a proxy.

leonardotc:
This is outdated according to @b-garbacz

PR #386, opened 2025-11-21
Refactor validator picking logic

jusufsuljic:


src/core/network/services/ConnectionManager.js
```diff
@@ -1,121 +1,164 @@
...
-                this.send(message, retries - 1)
+            entry.connection.messenger.send(message);
+            entry.sent = (entry.sent || 0) + 1;
+            if (entry.sent >= this.#messageThreshold) {
+                const validatorAddress = PeerWallet.encodeBech32m(
+                    TRAC_NETWORK_MSB_MAINNET_PREFIX, // TODO: This won't work for other networks. Make it dynamic.
+                    target
+                );
+                console.log("Removing validator after reaching message threshold:", validatorAddress);
+                this.remove(target); // TODO: In the future, only "flag for removal" and remove in a separate callback
```
b-garbacz: Observation:
There is no proper link to the ledger state. The connection manager does not check whether a new entry appears in the unsigned state after sending the data. The validator is removed right after calling `send` instead of waiting to confirm that the unsigned state was actually updated in the ledger.

We remove the node, but we do not check whether the transaction appeared in the unsigned state. In this case, shouldn’t we attempt the next try if the transaction failed for some reason? Should the responsibility for retries also be handled here?

PR #389, opened 2025-12-02
388 implement new endpoint that will allow to check nodes status

b-garbacz:
- Reworked acceptance test layout: one directory per endpoint under tests/acceptance/v1, with rpc.test.mjs as the entrypoint. Added new coverage (tx hashes, tx details, tx, broadcast,
    balance, account, etc.) and kept TODO/skip cases where handlers don’t yet validate.
  - New endpoint logic lives directly in the handler (agreed with Jusuf), not in the command layer.
  - CI workflow renamed to MSB-Unit-Tests, acceptance workflow now runs only after a successful unit workflow (or manually).

### DIFF
--- a/rpc/handlers.mjs
+++ b/rpc/handlers.mjs
@@ -1,22 +1,20 @@
 import { decodeBase64Payload, isBase64, sanitizeBulkPayloadsRequestBody, sanitizeTransferPayload, validatePayloadStructure } from "./utils/helpers.mjs"
 import { MAX_SIGNED_LENGTH, ZERO_WK } from "./constants.mjs";
+import { buildRequestUrl } from "./utils/URL.mjs";
 import { isHexString } from "../src/utils/helpers.js";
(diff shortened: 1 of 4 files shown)


rpc/handlers.mjs
```diff
@@ -250,3 +252,65 @@ export async function handleTransactionExtendedDetails({ msbInstance, respond, r
...
+            writingKey: nodeEntry.wk.toString('hex'),
+            isWhitelisted: nodeEntry.isWhitelisted,
+            isValidator: nodeEntry.isWriter,
+            isIndexer: nodeEntry.isIndexer,
+            license: licenseValue === 0n ? null : licenseValue.toString(),
+            balance: bufferToBigInt(nodeEntry.balance).toString(),
+            stakedBalance: bufferToBigInt(nodeEntry.stakedBalance).toString(),
+        });
+
+    } catch (error) {
```
leonardotc: This should be caught by the general handler.

jusufsuljic (approved):

PR #402, opened 2025-12-09
Remove function defaults

leonardotc:


rpc/handlers.mjs
```diff
@@ -258,7 +259,7 @@ export async function handleAccountDetails({ msbInstance, respond, req }) {
         return respond(400, { error: 'Parameter "confirmed" must be exactly "true" or "false"' });
     }
 
-    if (!isAddressValid(address)) {
+    if (!isAddressValid(address, TRAC_NETWORK_MSB_MAINNET_PREFIX)) {
```
leonardostsouza: Just an observation. All those `TRAC_NETWORK_MSB_MAINNET_PREFIX` arguments will need to be replaced by some other constant when the config PR comes online, right?

PR #410, opened 2025-12-18
MSB - Interactive mode and RPC code refactoring - pt1

jusufsuljic:
Next high‑value step: introduce a small src/interface/msbInterface.js that exposes data‑oriented operations (balance, account, tx details, broadcast, etc.) and refactor rpc_services.js to be a thin pass‑through to that instead of using cli.js helpers or state directly.

Then: point the CLI adapter (CommandLineHandler) at the same interface module for read‑type operations (getBalance, account, tx details), while keeping admin/role operations as direct MSB method calls for now.

Finally (optional): add a few focused tests around the CLI adapter using a mocked interface/MSB to validate parsing and routing.


leonardotc:
This was reused by my pr/

PR #432, opened 2026-01-08
fix: resolve string treated as buffer without explicit conversion

leonardostsouza:


src/core/network/messaging/validators/base/BaseResponse.js
```diff
@@ -71,8 +71,7 @@ class BaseResponse {
 
                 break;
             default:
-                const addressString = bufferToAddress(message.address, this.#config.addressPrefix);
-                publicKey = PeerWallet.decodeBech32m(addressString);
+                publicKey = PeerWallet.decodeBech32m(message.address);
```
b-garbacz: This looks ok and makes sense

src/core/network/messaging/validators/base/BaseResponse.js
```diff
@@ -81,7 +80,8 @@ class BaseResponse {
 
         const messageWithoutSig = { ...message };
         delete messageWithoutSig.sig;
-        const hash = await PeerWallet.blake3(JSON.stringify(messageWithoutSig));
+        const hashInput = b4a.from(JSON.stringify(messageWithoutSig), 'utf8');
```
b-garbacz: We can start the discussion here. In `src/core/network/messaging/handlers/GetRequestHandler.js`, the message is built without `UTF‑8` encoding, while here you already apply that encoding, which will cause inconsistency.


1. The question is: it must have been working in the version where you hadn’t yet converted messageWithoutSig to binary with encoding. If it didn’t work, then test it. But I believe that it was working, nodes find each other because, if they do, it means that the signatures are valid.
2. What if some nodes use b4a.from(JSON.stringify(messageWithoutSig), 'utf8') and others still use the previous version await PeerWallet.blake3(JSON.stringify(messageWithoutSig));? The messages will become incompatible.

leonardostsouza: Yes. Lots of error messages like

```sh
NetworkMessages: Failed to handle incoming message: Failed to route message: Invalid input: must be a Buffer or Uint8Array. Pubkey of requester is ...
```

Then I investigated and found these parts in the code as the source of the problem

PR #469, opened 2026-02-11
feat: remove constants in favor of config

leonardotc:
### What this does?

- It centralizes configuration of things that are environment-related (aka stuff that might change from env to env). That is tied to the goal of having the configuration as the only source of truth for that stuff (all those options will die eventually... unless its a param that needs to be injected by the docker etc). 
- It normalizes the test to use as much from the configuration as possible.
- Uses development config for tests.
- Creates some testnet placeholders (and some real ones)

There is also the side-effect on the fixtures etc. I did that to verify the system would work with a different prefix, network id etc. Tests were green. Some of the changes are reflected here (they can be rolled back).

### DIFF
--- a/tests/unit/messages/network/NetworkMessageBuilder.test.js
+++ b/tests/unit/messages/network/NetworkMessageBuilder.test.js
@@ -1,7 +1,6 @@
 import { test } from 'brittle';
 import b4a from 'b4a';
 import PeerWallet from 'trac-wallet';
-import { TRAC_NETWORK_MSB_MAINNET_PREFIX } from 'trac-wallet/constants.js';
 import { v7 as uuidv7 } from 'uuid';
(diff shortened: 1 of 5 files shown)


src/config/env.js
```diff
@@ -1,15 +1,51 @@
...
     MAINNET: 'mainnet',
     DEVELOPMENT: 'development',
     TESTNET1: 'testnet1'
 }
 // TODO: CREATE TEST ENV CONFIG SIMILAR TO MAINNET AND USE IT IN TESTS.
-// TODO: CREATE TESTNET1 ENV CONFIG and update npm scripts to run node witn mainnet or testnet1.
 
 const configData = {
+    [ENV.TESTNET1]: {
+        addressLength: 67,
```
b-garbacz: Length is 67 like for mainnet?

b-garbacz: Yes, you are totally right. My bad!

PR #471, opened 2026-02-12
feat: Implement Proof Of Publication

b-garbacz:


src/utils/deepEqualApplyPayload.js
```diff
@@ -0,0 +1,43 @@
...
+
+export const isDeepEqualApplyPayload = (left, right) => {
+    if (left === right) return true;
+
+    const leftIsBuffer = b4a.isBuffer(left);
+    const rightIsBuffer = b4a.isBuffer(right);
+
+    if (leftIsBuffer || rightIsBuffer) {
+        if (!leftIsBuffer || !rightIsBuffer) return false;
+        return b4a.equals(left, right);
```
leonardostsouza: It seems line 12 is too complex for what it is doing and maybe unnecessary. Won't the `return b4a.equals(left, right)` solve this case already?

src/core/network/protocols/legacy/handlers/LegacyRoleOperationHandler.js
leonardostsouza: `connection` is not used in the function

src/utils/deepEqualApplyPayload.js
b-garbacz: No, we need it. In this case for deep equality we want to receive false, instead of exception. But I can simplify this line to:
```
if (leftIsBuffer || rightIsBuffer) {
    return leftIsBuffer && rightIsBuffer && b4a.equals(left, right);
  }
```
Because this is this same right?

src/core/network/protocols/legacy/handlers/LegacyRoleOperationHandler.js
b-garbacz: Fixed

PR #489, opened 2026-02-24
feat: cli refactor

leonardotc:


leonardotc:
There is actually one extra bit that would be required by trac-peer which is bubbling up core events.

```
async waitForSignedLengthAtLeast(targetSignedLength) {
    while (this.#msb.state.base.view.core.signedLength < targetSignedLength) {
        await new Promise((resolve) => this.#msb.state.base.view.core.once('append', resolve));
    }
}
```

As a general rule, this kind of coupling between the projects aint ideal. I am sort of working to remove it but that will require some experimentation. On peer, we add transactions to a pool (right after we broadcast through MSB). In peer, we process from that pool and append to its own base. During that process we need to await msb signedState (which we do during processing). We do that because there is a race condition between the signedState and the observer. That has two side effects: 

- the architecture leakage that exposes msb internals to peer
- the wait (its async await so it doesnt bar the thread) during tx processing which is general not particularly good

There is going to be a PR on peer to bump msb to latest version and to match the signed state during the observer poll. This has to be tested as it will increase the pool size and cause a mismatch with the txDelay (which is parameterized in peer).

jusufsuljic (approved):
Is there a practical need for unit tests around this cli? Like something that can assure us nobody made a breaking change in that field? Otherwise approved.

leonardotc:
I think for the cli, the ideal would be more integrated tests, the problem is that the cli interface is not structured enough to test deterministically without "spagetting it up".

jusufsuljic (approved):

PR #498, opened 2026-02-26
feat: define non overridable store names

leonardotc:


src/utils/type.js
leonardostsouza: This commentary is wrong. The function doesn't returns `false` for `null`

PR #513, opened 2026-03-02
refactor: remove static PeerWallet calls

leonardotc:
The purpose here is to swap all the proxy PeerWallet calls for trac-crypto-api calls. The benefits are

- Having a single point of change in case of interface change/deprecation.
- Eliminate indirection (and therefore complexity).
- Organizational segregation between stateful and stateless calls.
- Single Responsibility principle on the Wallet side
- Enable easiness to restructure the wallet project

Out of scope:
- NetworkWalletFactory was not changed because it will be eliminated altogether by responsibilities that will be absorbed by the wallet.

### DIFF
--- a/src/core/network/protocols/shared/validators/PartialOperationValidator.js
+++ b/src/core/network/protocols/shared/validators/PartialOperationValidator.js
@@ -156,7 +156,7 @@ class PartialOperationValidator {
             );
         }
 
-        if (!verify(incomingSignature, messageHash, incomingPublicKey)) {
+        if (!tracCryptoApi.signature.verify(incomingSignature, messageHash, incomingPublicKey)) {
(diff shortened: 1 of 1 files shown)


src/core/network/protocols/shared/validators/PartialOperationValidator.js
```diff
@@ -78,7 +78,7 @@ class PartialOperationValidator {
             throw new SharedValidatorRejectionError(ResultCode.REQUESTER_ADDRESS_INVALID, 'Invalid requesting address in payload.');
         }
 
-        const incomingPublicKey = PeerWallet.decodeBech32mSafe(incomingAddress);
+        const incomingPublicKey = decodeBech32mSafe(incomingAddress);
```
leonardostsouza: Why not use `tracCryptoApi.address.decodeSafe(address)` ?

PR #516, opened 2026-03-04
fix(network): resolve validator selection bias and optimize observer …

lucasfernandes:
## The Problem
A reported visual bug (validator count not decreasing on `/remove_writer`) revealed a critical Sybil attack vector. Malicious nodes could spam `/add_writer` to duplicate their ledger entries, bypassing uniform randomness and skewing network selection probabilities in their favor.

## The Solution
Instead of requiring a hardfork to alter the ledger structure, the selection logic was isolated and completely overhauled in the `ValidatorObserverService` using a reactive, high-performance architecture:

1. **Fair Selection & O(1) Memory Pool:** Implemented an encapsulated Array + Hash Map structure using the Swap and Pop technique. This deduplicates entries, ensures fair probability, and guarantees O(1) performance for pool manipulation. Bound by a `MAX_POOL_SIZE` limit.
2. **Event-Driven Architecture (Real-Time):** The Observer now reacts instantly to `State` events (`WRITABLE`, `UNWRITABLE`, `IS_INDEXER`). Event payloads now emit `{ wk, publicKey }` objects, allowing the pool to update in O(1) without breaking legacy network connections.
3. **Zero-I/O Fallback (Atomic Swap):** Completely removed the heavy Hyperbee DB polling and chunking. The periodic sync fallback now iterates natively over Autobase's in-memory `system.list()`, performing an atomic swap of the pool with zero disk I/O.

## Tests & Coverage
- **Autobase Native Mocks:** Upgraded the test suite to simulate the asynchronous Autobase native iterator (`async function*`).
- **Behavioral Assertions:** Replaced brittle mathematical assertions with robust integration tests validating connection attempts, connection drops, and pool state recovery.
- **Coverage & Safety:** Reached **>92%** statement coverage. Fully tests boundary conditions (admin-indexer limits, DB crash recovery) while ensuring strict `PeerWallet` mock isolation via `setup/teardown` hooks.


src/core/network/services/ValidatorObserverService.js
```diff
@@ -26,6 +26,10 @@ class ValidatorObserverService {
     #scheduler;
     #address;
     #isInterrupted
+    #activeWriterArray = [];
+    #addressIndex = new Map();
```
b-garbacz: I would recommend to limit size of this map to 10k addresses

PR #519, opened 2026-03-06
Fix validator selection bias and introduce event-driven writer pool

lucasfernandes:
# PR Description: Native Validator Selection (Autobase Integration)

## The Problem
The previous implementation relied on redundant state tracking and manual event listeners, which caused state desync and potential Sybil-style bias (duplicate entries in the writer list).

## The Solution: "Native System View" Integration
Redesigned the `ValidatorObserverService` to treat the **Autobase System List** as the only Source of Truth, eliminating redundant I/O and in-memory state tracking.

### Key Changes:
* **Native SSoT:** Consumes `system.list()` directly from Autobase memory. Removed all manual arrays, Maps, and event listeners (`WRITABLE`, etc.).
* **Automatic Deduplication:** Naturally filters `isRemoved` and `isIndexer` flags provided by the native system view.
* **Zero DB Overhead:** Optimized performance by trusting the system list roles, removing redundant `getNodeEntry` queries to the ledger.
* **Memory Efficiency:** Switched to a stateless architecture where the selection pool is transient and garbage-collected every cycle.
* **Fair Selection:** Implemented randomized candidate selection directly from the native stream once per worker loop.

## Tests & Coverage
The test suite was fully refactored to validate the new stateless architecture.

* **Lines:** 100% (80/80)
* **Functions:** 100% (7/7)
* **Branches:** 92%

Verified scenarios include: filtering self-address, skipping removed nodes/indexers, lifecycle safety (start/stop), and configuration compliance.

### DIFF
--- a/src/core/network/services/ValidatorObserverService.js
+++ b/src/core/network/services/ValidatorObserverService.js
@@ -1,177 +1,116 @@
-import PeerWallet from "trac-wallet";
-import b4a from "b4a";
-import { bufferToAddress } from '../../state/utils/address.js';
-import { sleep } from '../../../utils/helpers.js';
-import Scheduler from "../../../utils/Scheduler.js";
-import Network from "../Network.js";
+import { bufferToAddress } from '../../../core/state/utils/address.js';
+import Scheduler from '../../../utils/Scheduler.js';
 import { Logger } from '../../../utils/logger.js';
 
-const DELAY_INTERVAL = 50
-const VALIDATOR_CANDIDATES_PER_CYCLE = 10
-const POLL_INTERVAL = (VALIDATOR_CANDIDATES_PER_CYCLE + 1) * DELAY_INTERVAL // This is to avoid more than one instance of the worker running at the same time
+const DEFAULT_POLL_INTERVAL = 5000;
 
 class ValidatorObserverService {
-    #config;
-    #state;
     #network;
+    #state;
+    #selfAddress;
+    #config;
     #scheduler;
-    #address;
-    #isInterrupted
     #logger;
 
-    /**
-     * @param {Network} network
-     * @param {State} state
-     * @param {string} address
-     * @param {Config} config
-     **/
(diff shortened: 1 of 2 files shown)


src/core/network/services/ValidatorObserverService.js
```diff
@@ -1,187 +1,103 @@
...
-                this.#network.validatorConnectionManager.remove(validatorPubKeyBuffer)
-            }
-        }
-
-        // Connection validation rules:
-        // - Cannot connect if already connected to a validator
-        // - Validator must exist and be a writer
-        // - Cannot connect to indexers, except for admin-indexer
-        // - Admin-indexer connection is allowed only when writers length has less than 10 writers
-        if (this.#network.validatorConnectionManager.connected(validatorPubKeyBuffer) ||
```
b-garbacz: What is the reason for deleting this check?

src/core/network/services/ValidatorObserverService.js
```diff
@@ -1,187 +1,103 @@
...
-                this.reachedMax = true;
-                debugLog('Max validator connections reached. Skipping this cycle.');
-                const now = Date.now();
-                const elapsed = now - this.initTimestamp;
-                debugLog('>>> Time elapsed since start (ms):', elapsed);
-            }
+            // Standard delay between cycles to keep CPU usage low
+            await sleep(interval);
         }
-        next(POLL_INTERVAL);
```
b-garbacz: What is the reason for deleting this next() and replaced this with sleep?

lucasfernandes: I used sleep to keep the async worker loop simple, but I see your point about the Scheduler. I'll handle it.

src/core/network/services/ValidatorObserverService.js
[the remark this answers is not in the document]
lucasfernandes: Restored in the latest refactor to align with protocol rules

PR #546, opened 2026-04-07
feat: add eslint 

leonardotc:


leonardotc:
There are three commits. The middle one is just a bunch of autofixes and some manual ones (but swapping exports order etc).

b-garbacz (approved):

PR #547, opened 2026-04-07
fix: intermittent broadcast failures and refactor active…

### LINKED ISSUE #545: bug(network): intermittent broadcast failures during CLI add/remove writer commands
## Context
*Note: This issue was identified and isolated during the review process of the Validator Observer refactor in PR #542.*

While testing the main branch and validating network changes, it was observed that executing `/add_writer` and `/remove_writer` via the CLI occasionally triggers a `Failed to broadcast transaction. Try again later.` error.

## Current Behavior
The CLI command logs indicate that the system attempts to hit the validator 3 times, failing the first 2 attempts before eventually throwing the broadcast error. This suggests a potential delay or failure in peer propagation during membership transitions.

## Expected Behavior
The broadcast should execute cleanly on the first attempt without throwing errors or requiring multiple retries when adding or removing a writer.

## Action Items
- [ ] Investigate the broadcast flow inside the network orchestrator during state changes.
- [ ] Check how `ConnectionManager` and `Network.js` handle live vs. dead writers when the CLI triggers a membership update.
- [ ] Adjust the underlying retry/timeout logic to prevent these CLI broadcast failures.

lucasfernandes:
## Description
  We were experiencing intermittent broadcast failures during the `add` and `remove` writer CLI commands. The root cause was the node exceeding the
  network connection limit (`maxWritersForAdminIndexerConnection`). The active validator count was incorrect as it included the Admin node itself and
   previously removed validators. This blocked the Admin from validating and caused network congestion.

  ## What was done
  * **State.js Fix:** Implemented `getActiveWriterCount` to correctly filter and count only actual active validators. Added result caching keyed on
  `system.core.length` to prevent stale reads during concurrent add/remove writer operations. Added `isAdminAllowedToValidate` to strictly enforce
  the connection limit.
  * **Network Cleanup (index.js):** Removed duplicate broadcast logic and premature disconnect events that were killing the node unexpectedly.
  * **Config Adjustments:** Fixed the bootstrap hash configuration.
  * **Tests:** Fixed test pollution (state leakage) in the `broadcastTransactionResponse` test suite. Added unit tests fully compatible with the
  Holepunch/Bare runtime for validator counting logic. Added integration tests validating that `system.core.length` correctly triggers cache
  invalidation on a real Autobase instance.

### DIFF
--- a/src/config/env.js
+++ b/src/config/env.js
@@ -114,7 +114,7 @@ const configData = {
         addressPrefix: TRAC_NETWORK_MSB_MAINNET_PREFIX,
         addressPrefixLength: TRAC_NETWORK_MSB_MAINNET_PREFIX.length,
         bech32mHrpLength: TRAC_NETWORK_MSB_MAINNET_PREFIX.length + 1, // len(addressPrefix + separator)
-        bootstrap: '12f7f1668eac2e691e17cbc6a53e509c5cee78cdcac562313091c64e5fd077d6',
+        bootstrap: '6d2ffdacc7c1813f447baae3b43207437c469472b5a92de45ebec078c16713ea',
         channel: '12312313123123',
         dhtBootstrap: ['116.202.214.149:10001','157.180.12.214:10001','node1.hyperdht.org:49737','node2.hyperdht.org:49737','node3.hyperdht.org:49737'],
         derivationPath: address.MAINNET_DERIVATION_PATH,
--- a/src/config/env.js
+++ b/src/config/env.js
@@ -114,7 +114,7 @@ const configData = {
         addressPrefix: TRAC_NETWORK_MSB_MAINNET_PREFIX,
         addressPrefixLength: TRAC_NETWORK_MSB_MAINNET_PREFIX.length,
         bech32mHrpLength: TRAC_NETWORK_MSB_MAINNET_PREFIX.length + 1, // len(addressPrefix + separator)
-        bootstrap: '6d2ffdacc7c1813f447baae3b43207437c469472b5a92de45ebec078c16713ea',
+        bootstrap: '12f7f1668eac2e691e17cbc6a53e509c5cee78cdcac562313091c64e5fd077d6',
         channel: '12312313123123',
         dhtBootstrap: ['116.202.214.149:10001','157.180.12.214:10001','node1.hyperdht.org:49737','node2.hyperdht.org:49737','node3.hyperdht.org:49737'],
         derivationPath: address.MAINNET_DERIVATION_PATH,
--- a/src/core/network/Network.js
+++ b/src/core/network/Network.js
@@ -255,6 +255,26 @@ class Network extends ReadyResource {
         return this.#pendingConnections.size;
     }
 
+    disconnectValidatorPeer(publicKey, reason = 'validator peer invalidated') {
+        const publicKeyHex = this.#normalizePublicKey(publicKey);
+        if (!publicKeyHex) return false;
+
+        const hadPendingValidatorConnection = this.#clearPendingValidatorConnection(publicKeyHex);
+        const isTrackedValidator = this.#validatorConnectionManager.exists(publicKeyHex);
+
+        if (isTrackedValidator) {
+            this.#logger.debug(`Network.disconnectValidatorPeer: detaching tracked validator ${publicKeyHex}. Reason: ${reason}`);
+            this.#validatorConnectionManager.remove(publicKeyHex, { endConnection: false });
+        }
+
+        if (hadPendingValidatorConnection && this.#swarm?.peers?.has(publicKeyHex)) {
+            this.#logger.debug(`Network.disconnectValidatorPeer: leaving peer ${publicKeyHex}. Reason: ${reason}`);
+            this.#swarm.leavePeer(b4a.from(publicKeyHex, 'hex'));
+        }
+
+        return hadPendingValidatorConnection || isTrackedValidator;
+    }
+
     async #getOrGenerateWallet(store, wallet) {
         if (!this.#config.enableWallet) {
             const keyPair = await store.createKeyPair(TRAC_NAMESPACE);
@@ -304,6 +324,23 @@ class Network extends ReadyResource {
         if (!this.#pendingConnections.has(publicKey)) return;
         this.emit(EventType.VALIDATOR_CONNECTION_READY, { publicKey, type, connection });
         this.#logger.debug(`Network.finalizeConnection: Connected to peer: ${publicKey} as type: ${type}`);
+    }
+
+    #normalizePublicKey(publicKey) {
+        if (typeof publicKey === 'string') return publicKey;
+        if (b4a.isBuffer(publicKey)) return b4a.toString(publicKey, 'hex');
+        return null;
+    }
+
+    #clearPendingValidatorConnection(publicKeyHex) {
+        if (!this.#pendingConnections.has(publicKeyHex)) return false;
+
+        const { timeoutId, type } = this.#pendingConnections.get(publicKeyHex);
+        if (type !== 'validator') return false;
+
+        clearTimeout(timeoutId);
+        this.#pendingConnections.delete(publicKeyHex);
+        return true;
     }
 }
 
--- a/src/core/network/Network.js
+++ b/src/core/network/Network.js
@@ -267,7 +267,9 @@ class Network extends ReadyResource {
             this.#validatorConnectionManager.remove(publicKeyHex, { endConnection: false });
         }
 
-        if (hadPendingValidatorConnection && this.#swarm?.peers?.has(publicKeyHex)) {
+        const shouldLeavePeer = hadPendingValidatorConnection || isTrackedValidator;
+
+        if (shouldLeavePeer && this.#swarm?.peers?.has(publicKeyHex)) {
             this.#logger.debug(`Network.disconnectValidatorPeer: leaving peer ${publicKeyHex}. Reason: ${reason}`);
             this.#swarm.leavePeer(b4a.from(publicKeyHex, 'hex'));
         }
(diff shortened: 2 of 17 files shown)


src/core/state/State.js
```diff
@@ -177,9 +177,12 @@ class State extends ReadyResource {
     }
 
     async isAdminAllowedToValidate() {
+        if (!this.writingKey || !this.#config?.bootstrap) return false;
```
leonardotc: You should rely that there is a bootstrap on the config.

src/index.js
```diff
@@ -199,17 +196,6 @@ export class MainSettlementBus extends ReadyResource {
     }
 
     async #stateEventsListener() {
-        this.#state.on(CustomEventType.IS_INDEXER, (publicKey) => {
-            if (this.#network.validatorConnectionManager.exists(publicKey)) {
-                this.#network.validatorConnectionManager.remove(publicKey)
-            }
-        })
```
b-garbacz, line 212: Question - why are we deleting this?

src/core/state/State.js
```diff
@@ -193,6 +196,35 @@ class State extends ReadyResource {
         return Object.values(this.#base.system.indexers);
     }
 
+    async getActiveWriterCount({ excludeAdmin = false } = {}) {
```
b-garbacz: Just a detail and this is not important at all - What do you think to instead making an object like this `{ excludeAdmin = false } = {}`, we could just provide `excludeAdmin = false` as default value. Then in `await this.getActiveWriterCount(true);` you can just provide value. Depends from you.

leonardostsouza: Just an observation. This function is I/O-heavy. Fortunately it is only invoked inside `isAdminAllowedToValidate()`. 

`isAdminAllowedToValidate()` is invoked in several places, however. I would recommend we cache the result of this function in a constant instead of executing this several times during initialisation.

tests/unit/state/stateModule.test.js
```diff
@@ -15,8 +15,8 @@ async function runStateTests() {
     await import('./apply/state.apply.test.js');
     if (!isBare()) {
         await import('./State.test.js');
+        await import('./stateAdminValidation.test.js');
```
leonardostsouza, line 17: Does it still make sense to exclude these tests in Bare env?

src/index.js
lucasfernandes: The listener made sense, but the previous implementation was too shallow: it only interacted with the validator connection manager and did not handle pending validator attempts or centralize the invalidation rule in the right place.

I addressed it as follows:
1. I reintroduced the listeners in `src/index.js`, but they now only delegate to `this.#network.disconnectValidatorPeer(...)`.
2. I added `disconnectValidatorPeer()` in `src/core/network/Network.js`, which clears pending validator connection attempts and detaches tracked validators from the validator pool in a centralized way.
3. I kept the shutdown guard in the listeners so this does not run while the node is closing.

One important detail: when a validator is invalidated by a state event, we now remove it from the validator pool without force-closing the active socket, so we do not interrupt in-flight broadcast responses.

src/core/state/State.js
```diff
@@ -208,6 +212,44 @@ class State extends ReadyResource {
...
+
+                // Non-admin indexers do not participate in validator capacity decisions.
+                if (value.isIndexer && address !== adminAddress) continue;
+                if (excludeAdmin && address === adminAddress) continue;
+
+                activeAddresses.add(address);
+            }
+        } catch (error) {
+            this.#safeLogApply("ActiveWriterCount", error?.message ?? "Failed to scan active writers");
+            return 0;
```
leonardostsouza: Returning `0` here is not a good choice. This would make `isAdminAllowedToValidate` return true on exception, and the error would not be handled properly.

src/core/network/services/MessageOrchestrator.js
```diff
@@ -168,6 +177,30 @@ class MessageOrchestrator {
...
 
+    async #isIdempotentSuccess(resultCode, message) {
+        if (!this.#idempotentSuccessCodes.has(resultCode)) return false;
+
+        const txHash = this.#extractTxHash(message);
+        if (!txHash) return false;
+
+        // A short wait covers the race where a first validator committed the tx
+        // and a retried validator only observes it as "already exists/completed".
+        const timeout = Math.min(this.#config.messageValidatorResponseTimeout ?? 2000, 2000);
```
leonardostsouza: We are ignoring configured timeout to force a hard-coded 2 seconds if `messageValidatorResponseTimeout ` is bigger than that. Is this intentional?

tests/unit/state/stateModule.test.js
```diff
@@ -13,11 +13,11 @@ async function runStateTests() {
     // These tests are skipped temoporarily because the mock library sinon does not work with bare.
     // TODO: replace esmock, sinon is actually fine
     await import('./apply/state.apply.test.js');
+    await import('./stateAdminValidation.test.js');
     if (!isBare()) {
```
leonardostsouza, line 16: `stateAdminValidation.test.js` imports dependencies that are bare-sensitive (e.g.: fs, Corestore).

I recommend importing this after the `isBare` guard

src/config/env.js
```diff
@@ -114,7 +114,7 @@ const configData = {
         addressPrefix: TRAC_NETWORK_MSB_MAINNET_PREFIX,
         addressPrefixLength: TRAC_NETWORK_MSB_MAINNET_PREFIX.length,
         bech32mHrpLength: TRAC_NETWORK_MSB_MAINNET_PREFIX.length + 1, // len(addressPrefix + separator)
-        bootstrap: '12f7f1668eac2e691e17cbc6a53e509c5cee78cdcac562313091c64e5fd077d6',
+        bootstrap: '6d2ffdacc7c1813f447baae3b43207437c469472b5a92de45ebec078c16713ea',
```
leonardostsouza: Looks like this is leftover test data. Please, revert

src/core/network/services/MessageOrchestrator.js
```diff
@@ -168,6 +177,30 @@ class MessageOrchestrator {
         return success;
     }
 
+    async #isIdempotentSuccess(resultCode, message) {
```
b-garbacz, line 192: Please add a comment here explaining the retry/idempotency case in `/* */` commentary block for js docs. 

I noticed that you have added this because this path is needed when a previous validator already committed the tx but the requester did not accept the response, so a retry may return `TX_ALREADY_EXISTS` or `OPERATION_ALREADY_COMPLETED`. The local unsigned state check is what makes treating that as success safe. However, other devs should understand why this is important :P

src/core/network/Network.js
```diff
@@ -255,6 +255,28 @@ class Network extends ReadyResource {
...
+        if (isTrackedValidator) {
+            this.#logger.debug(`Network.disconnectValidatorPeer: detaching tracked validator ${publicKeyHex}. Reason: ${reason}`);
+            this.#validatorConnectionManager.remove(publicKeyHex, { endConnection: false });
+        }
+
+        const shouldLeavePeer = hadPendingValidatorConnection || isTrackedValidator;
+
+        if (shouldLeavePeer && this.#swarm?.peers?.has(publicKeyHex)) {
+            this.#logger.debug(`Network.disconnectValidatorPeer: leaving peer ${publicKeyHex}. Reason: ${reason}`);
+            this.#swarm.leavePeer(b4a.from(publicKeyHex, 'hex'));
```
b-garbacz, line 269: This is just an observation and we can start discussion here. Why are we using leavePeer? The reason is that we want to stop maintaining a direct connection and stop trying to reconnect with a specific peer.

Do you think it makes sense to first execute leavePeer and only then remove it from the connection manager?

First we terminate the actual connection, and only after that we remove it from the manager which is responsible for managing validator connections.

src/core/state/State.js
```diff
@@ -208,6 +212,44 @@ class State extends ReadyResource {
...
+                if (!address) continue;
+
+                // Non-admin indexers do not participate in validator capacity decisions.
+                if (value.isIndexer && address !== adminAddress) continue;
+                if (excludeAdmin && address === adminAddress) continue;
+
+                activeAddresses.add(address);
+            }
+        } catch (error) {
+            this.#safeLogApply("ActiveWriterCount", error?.message ?? "Failed to scan active writers");
```
b-garbacz: Observation - `#safeLogApply` seems to be used mainly in the apply/contract validation flow.

  `getActiveWriterCount()` looks more like a runtime helper used by `isAdminAllowedToValidate()`, so I wonder if logging through the apply logger is the best fit here.

  Maybe we could either remove this `try/catch` and let the error propagate naturally, or use the regular logger if we want local visibility in this method? (to be discussed of course, maybe I am wrong.)

  Could you also check what happens when `isAdminAllowedToValidate()` throws? It is used from `V1BroadcastTransactionOperationHandler`, `BaseStateOperationHandler`, and `TransactionPoolService`, so I just want to make sure an exception here is handled as expected.

src/core/network/services/MessageOrchestrator.js
lucasfernandes: Oki, I'm going step by step here. The first one: jsdoc added.

src/core/state/State.js
lucasfernandes: You're correct here. The #safeLogApply doesn't belong here. getActiveWriterCount is a runtime helper, not part of the apply/contract validation flow.
  
I traced all three callers of isAdminAllowedToValidate: V1BroadcastTransactionOperationHandler, BaseStateOperationHandler, and TransactionPoolService. All of them already have error handling upstream that covers this path, so it's safe to let the error propagate naturally.

Removed the try/catch entirely.

src/core/network/Network.js
lucasfernandes: The order you suggested look more natural, but it won't affect the result. Reordered. 
`leavePeer` now runs before remove.

src/core/state/State.js
lucasfernandes: Done, already simplified

b-garbacz (approved):
LGTM

PR #553, opened 2026-04-10
chore: centralize v1 protocol error codes and connection closing policies

### LINKED ISSUE #520: Refactor: centralize ResultCode/V1ProtocolError policies
# Requisites
- The **EPIC v1** must be merged before starting this refactor.

# Description 
Okay, so the plan is to create a new module that defines the behavior based on `V1ProtocolError`, in a similar way to `ResultCodePolicy.js` It would be even better to rename that file to `ResultCodesPolicies.js`.

In that module, we would have:
- a sender connection handling policy (Success, Rotate, NoRotate)
- a validator connection handling policy (the ResultCodes that should cause us to close the validator connection).

It will have one big benefit - centralizing all policies into a single file.

@leonardotc @jusufsuljic @leonardostsouza I’m going to merge the current PR as is, and then create a followup issue to refactor it in this direction.

_Originally posted by @b-garbacz in https://github.com/Trac-Systems/main_settlement_bus/pull/514#discussion_r2896638405_

b-garbacz:
## Summary

  This PR simplifies V1 network error handling and makes the connection policy easier to understand.

  Before, the behavior was split between many custom error classes, endConnection flags, and handler specific logic.
  Now, ResultCode is the main place that defines protocol behavior and connection policy.

  ## What Changed

  - replaced scoped V1 errors and shared validator errors with V1ProtocolError(resultCode, message)
  - removed endConnection from protocol errors
  - moved validator connection closing decisions to connectionPolicies.shouldEndConnection(resultCode)
  - kept sender retry and rotation logic in connectionPolicies.resultToValidatorAction(resultCode)
  - made validator connection closing behavior explicit for codes such as INVALID_PAYLOAD, TX_INVALID_PAYLOAD, and shared validator request payload validation errors
  - kept shared validator handling aligned with the centralized policy
  - simplified sender timeout and fallback error handling in PendingRequestService
  - added documentation for all V1 ResultCode values in docs/v1_result_code_policies.md
  - added a docs checker script that verifies the documentation is in sync with protobuf enums, ResultCode constants, and current connection policies
  - added a docs workflow that runs the docs checker on pull requests
  - updated and expanded tests for connection policy, timeout handling, and broadcast transaction flows

  ## Why

  Before this change, it was hard to understand where transport behavior came from.
  Some decisions were inside error classes, some were in handlers, and some were controlled by flags.

  This change makes the code easier to read, easier to test, and easier to maintain.

  It also makes the protocol behavior more explicit, because the same ResultCode now defines:

  - sender validator rotation
  - validator connection closing
  - protocol level error reporting

  ## Behavioral Notes

  - shared validator result codes still close the connection on the validator when the policy says they should
  - validator connection closing rules are now defined in one place, including strict payload related codes like INVALID_PAYLOAD and TX_INVALID_PAYLOAD
  - TIMEOUT still rotates validators on the sender, but it does not close the connection on the validator after a response is sent
  - PendingRequestService internal rejected errors no longer need to be V1 protocol errors
  - the docs table now shows, for every ResultCode, whether the sender rotates the validator and whether the validator closes the connection
  - the docs checker now fails if a new protobuf ResultCode is added without updating src/utils/constants.js or the documentation


scripts/docs/check-v1-result-code-policies.js
```diff
@@ -0,0 +1,210 @@
...
+    return Object.entries(protobufResultCodeEnum)
+        .filter(([, value]) => typeof value === 'number')
+        .map(([name, value]) => ({
+            resultCodeName: name.replace(/^RESULT_CODE_/, ''),
+            value
+        }));
+}
+
+function parseBooleanMark(mark, columnName) {
+    if (mark === '✅') return true;
```
jusufsuljic: In my honest opinion these emojis gives off those "clearly vibe coded" vibes. Can we just use true and false, or 1s and 0s?

b-garbacz: I wanted to use emojis in this case, but I can change this to 0-1 to that the user doesn't feel this was vibe coded

b-garbacz: Done.

jusufsuljic (approved):

leonardostsouza (approved):
LGTM

PR #558, opened 2026-04-15
feat: docker publishing

### LINKED ISSUE #493: Enabling MSB node from README
# Description
1. Please see `README file`, and update it so that starting the node is easy in every scenario (mainnet/testnet, with/without RPC, env-based, Docker, etc.). If you want to introduce a consistent pattern that simplifies the startup process across all cases, feel free to do, any improvement is welcome.

2. Please make sure the `README` is updated accordingly (commands, examples, and expected behavior).

3. I would like to recommend to create a new task `Enabling MSB node from README`, to validate the documentation end-to-end. The goal is to be 100% sure that:
   - the documented commands actually work,
   - the built Docker image runs correctly,
   - users should be able to choose whether the service binds to `localhost` or `0.0.0.0`. When running inside a container, the node should bind to `0.0.0.0` by default so it is reachable from the host through published ports. On the host side, exposure should remain configurable (for example, publish the port on `127.0.0.1/localhost` for local only access, or on `0.0.0.0` to expose it externally). But again user should be aware what it is selecting.

_Originally posted by @b-garbacz in https://github.com/Trac-Systems/main_settlement_bus/pull/488#discussion_r2849119739_
 
### Acceptance Criteria
- [x] All README commands work end-to-end (host + Docker).
- [x] Docker image builds and runs successfully.
- [x] If something does not work, then update README to the working version of the command/instruction.

leonardotc:


leonardotc:
This is synced with the linter. That needs to be merged first.

PR #573, opened 2026-05-11
feat: bootstrap consesus algorithm

### LINKED ISSUE #569: Prototype the consensus algorithm
Assume this will be performed by the admin.

Admin should collect signatures from every participant in the consensus to decide on an epoch. The datastructure that will be appended should be in binary format. The nonce and message should be the same for every node and the signatures field should contain the signature and message.

```
{
  "previousEpoc": "1213",
  "nonce": 12312312
  "signatures": [{
     "address": "",
    "signature": """
  }]
}
```

For every signature, it is the role of the (ADMIN) to collect it for that nonce + message.

leonardotc:


proto/consensus/v1/messages/epoch_proof_proposal_request.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package network.v1;
+
+message EpochProofProposalData {
+  uint32 protocol_version = 1;
+  uint64 epoch = 2;
+  bytes previous_epoch_hash = 3;
+  bytes committed_hash = 4;
+  bytes leader_id = 5;
```
b-garbacz, line 10: why leader_id?

b-garbacz: Don't we have a designed data structures?

src/core/network/protocols/v1/validators/V1EpochProofProposalResponse.js
```diff
@@ -0,0 +1,58 @@
+import V1BaseOperation from "./V1BaseOperation.js";
+import b4a from "b4a";
+import tracCryptoApi from "trac-crypto-api";
+import { ResultCode, WRITER_BYTE_LENGTH } from "../../../../../utils/constants.js";
```
b-garbacz: Wasn’t the intention of the V1 protocol to focus on transactional communication?

Proof of time is a completely different protocol layer, and I think we should clearly separate it from the orchestrator and the ConnectionManager. These are responsible for finding validators.

Many things in this PR are nice, but too many concerns are mixed together. This will become painful very soon.

Let’s pause the implementation until we create the proper tasks first.

b-garbacz (changes requested):

b-garbacz (approved):

PR #610, opened 2026-06-08
feat: extend state apply builder to construct final epoch operation payload

### LINKED ISSUE #592: Extend state apply builder to construct final epoch operation payload

leonardostsouza:


b-garbacz (approved):
LGTM

PR #614, opened 2026-06-10
feat: implement base events for epoch service

lucasfernandes:


src/utils/constants.js
```diff
@@ -43,13 +43,11 @@ export const NetworkOperationType = Object.freeze({
     LIVENESS_RESPONSE: NetworkMessageType.MESSAGE_TYPE_LIVENESS_RESPONSE,
     BROADCAST_TRANSACTION_REQUEST: NetworkMessageType.MESSAGE_TYPE_BROADCAST_TRANSACTION_REQUEST,
     BROADCAST_TRANSACTION_RESPONSE: NetworkMessageType.MESSAGE_TYPE_BROADCAST_TRANSACTION_RESPONSE,
-    EPOCH_PROOF_PROPOSAL_REQUEST: NetworkMessageType.MESSAGE_TYPE_EPOCH_PROOF_PROPOSAL_REQUEST,
-    EPOCH_PROOF_PROPOSAL_RESPONSE: NetworkMessageType.MESSAGE_TYPE_EPOCH_PROOF_PROPOSAL_RESPONSE,
```
leonardostsouza, line 47: Both these constants are still being used in `ConsensusRouter.js` and `ConsensusValidationSchema.js`. 

Is this intentional?

src/core/state/State.js
```diff
@@ -129,10 +129,20 @@ class State extends ReadyResource {
...
+        if (!epochIdBuffer) return 0;
+
+        return lengthEntryUtils.decodeBE(epochIdBuffer)
+    }
+
     async currentEpoch() {
-        const latestEpochNumber = await this.getSigned(keys.CURRENT_INDEX)
-        return await this.getSigned(keys.EPOCH(latestEpochNumber))
+        const epochId = await this.currentEpochId()
+        return await this.getSigned(keys.EPOCH_DATA(epochId));
```
leonardostsouza, line 144: `epochProofData.keys` only exports `CURRENT_INDEX`, `EPOCH_HASH` and `EPOCH`. Maybe you forgot to change the definition in `epochProofData.js`?

src/core/network/Network.js
```diff
@@ -160,11 +170,27 @@ class Network extends ReadyResource {
...
+                this.#epochProofProposalService.stop();
+            }
+        });
     }
 
     cleanupNetworkListeners() {
         this.removeAllListeners(EventType.VALIDATOR_CONNECTION_TIMEOUT);
         this.removeAllListeners(EventType.VALIDATOR_CONNECTION_READY);
+        this.#state.removeAllListeners(CustomEventType.IS_INDEXER);
+        this.#state.removeAllListeners(CustomEventType.IS_NON_INDEXER);
```
leonardostsouza: I'm not sure it is a good idea to remove listeners from `state` from inside `network` module. Was this intentional?

src/core/network/Network.js
```diff
@@ -160,11 +170,27 @@ class Network extends ReadyResource {
             }
 
         });
+
+        this.#state.on(CustomEventType.IS_INDEXER, (bufferAddress) => {
+            const address = tracCryptoApi.address.encode(this.#config.addressPrefix, bufferAddress);
+            if (address === this.#wallet.address) {
+                this.#epochProofProposalService.start();
```
leonardostsouza, line 177: We should also check if the node is an admin. The admin should not run this service.

Also, is this not a duplication from line 105? Please, check

leonardostsouza (changes requested):

src/core/network/Network.js
leonardotc: In this case since its an event of becoming indexer, which the admin will never do. The major issue (which is incorrect) is that the indexers list returns the admin.

src/core/network/Network.js
leonardotc: It was intentional but I think this will actually break stuff. Iirc the main class subscribes to this event. I think this could be removed altogether on the state close.

src/utils/constants.js
leonardotc: It is deliberate and removing is correct because those are not network messages. We will removed from the other place (and use consensus). I do think this situation will be solved by a future pr for the schema validator.

src/core/network/Network.js
```diff
@@ -202,9 +226,11 @@ class Network extends ReadyResource {
             this.#validatorHealthCheckService = new ValidatorHealthCheckService(this.#config);
             await this.#validatorHealthCheckService.ready();
 
+            this.#consensusMessages = new ConsensusMessages(this.#state, this.#wallet, this.#config, this.#pendingRequestsService);
+
             this.#epochProofProposalService = new EpochProofProposalService(this.#state, this.#validatorConnectionManager, this.#wallet, this.#config);
             await this.#epochProofProposalService.ready();
-            this.#epochProofProposalService.start();
+            // this.#epochProofProposalService.start();
```
leonardostsouza, line 233: Please, remove commented-out code or explain when we should uncomment this in a commentary

leonardostsouza (approved):

PR #619, opened 2026-06-12
feat: extract scheduler

leonardotc:
The purpose of this is to extract a schedulable job logic into its own class. Define a clear interface so we just use public methods and unify the lifecycle methods of the services that operate with it. Ideally, they wouldnt "know" about this.#scheduler but rather the interface of the super class that uses it. This will offload stuff from the epoch service that, right now, is extremely ugly. This also proposes a common interface of a schedulable job as something that can be "open" and "closed", "stop" and "started". Maybe in the future we could even treat them as a composite in the network or state (lets not get into this rn).

### DIFF
--- a/src/core/consensus/services/EpochProofProposalService.js
+++ b/src/core/consensus/services/EpochProofProposalService.js
@@ -1,6 +1,6 @@
 import ReadyResource from 'ready-resource';
 import b4a from 'b4a';
-import Scheduler from '../../../utils/Scheduler.js';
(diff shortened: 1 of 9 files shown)


leonardotc:
Please, ignore the fact that `src/core/consensus/services/EpochProofProposalService.js` is still not using the refactor. That should be verified in the PR that implements that service.

lucasfernandes (approved):
LGTM

b-garbacz (approved):

PR #623, opened 2026-06-15
feat: node support for vdf service

lucasfernandes:
This pr adds node support to VDFService. Currently it only works on bare.

### DIFF
--- a/src/core/consensus/services/VDFNodeService.js
+++ b/src/core/consensus/services/VDFNodeService.js
@@ -1,37 +1,22 @@
 import ReadyResource from 'ready-resource';
 
-const isBare = typeof globalThis.Bare !== 'undefined';
-
-export class VDFService extends ReadyResource {
+export class VDFNodeService extends ReadyResource {
     #thread = null;
     #port = null;
 
     async _open() {
-        if (isBare) {
-            const { default: Channel } = await import('bare-channel');
-            const channel = new Channel();
-            this.#port = channel.connect();
-            this.#thread = new globalThis.Bare.Thread('./vdf-worker.js', { data: channel.handle });
-        } else {
-            const { Worker, MessageChannel } = await import('worker_threads');
-            const { port1, port2 } = new MessageChannel();
-            this.#port = this.#wrapNodePort(port1);
-            this.#thread = new Worker(new URL('./vdf-worker.js', import.meta.url), {
-                workerData: { port: port2 },
-                transferList: [port2],
-            });
-        }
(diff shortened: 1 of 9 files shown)


src/core/consensus/services/VDFNodeService.js
```diff
@@ -0,0 +1,57 @@
+import ReadyResource from 'ready-resource';
+
+export class VDFNodeService extends ReadyResource {
```
b-garbacz: VDFBareService and VDFNodeService duplicate the same calculateVDF() flow. This would fit a template method design (please see https://refactoring.guru/design-patterns/template-method). We could have one base class where VDFNode and VDFBare inherits some of methods. This base method could inherit by ready resource :)

src/core/consensus/services/VDFNodeService.js
```diff
@@ -0,0 +1,57 @@
...
+            transferList: [port2],
+        });
+    }
+
+    async _close() {
+        await this.#thread.terminate();
+        this.#port.close();
+    }
+
+    async calculateVDF(challenge, difficulty, discriminantSizeBits) {
```
b-garbacz: Hmnnn, shouldn’t this be managed by a queue and requestId?

  I know that a challenge will take a lot of time, but let’s consider an edge case where we start calculating 2 VDFs at the same time:
```
  const a = service.calculateVDF(challengeA, 100, 512);
  const b = service.calculateVDF(challengeB, 10, 512);
```
If these 2 calls run in parallel, both will send requests and both will start waiting for the “next” response from the port. The worker does not tag responses with any requestId, so the service does not know which response belongs to which promise.

The problem appears if request B finishes faster than request A. Then the first pending read() can receive response B and return it to caller A. So the VDF result for challengeB can end up resolving the promise for challengeA.

Maybe I am overthinking, but what do you think? It might be a problem??? Maybe we should check this. If in your opinion this case won't never happen then revolve this topic.

src/core/consensus/services/VDFNodeService.js
lucasfernandes: Ok, agreed. Implemented following this approach

src/core/consensus/services/VDFNodeService.js
lucasfernandes: Agreed. Added a serial queue (#queue) in the base to prevent response mismatches on concurrent calls to cover the race condition.

b-garbacz (approved):

PR #625, opened 2026-06-17
feat: Indexer messages and router

### LINKED ISSUE #628: Imeplement the router for consesus along with the required services (aka IndexerManager)
### Overview

Mechanical requirements:

- establish a mechanism for creating the protobuf communication channel
- **we would need to replicate the event logic that adds to the ConnectionManager for the consensus indexers**
- create a handler that does what the minion has to do to verify that the epoch generation process is correct, sign it, and send it back to the leader
- apply operation handler for the epoch (contract)

## Assumptions:

- we need to talk about the first epoch (genesis)
|> requirements

## Post-feature:
- do the QAs associated with the process
|> it has to work in a distributed way

leonardotc:
Things outside of scope:

Stuff associated with this refactor: https://github.com/Trac-Systems/main_settlement_bus/issues/631 and https://github.com/Trac-Systems/main_settlement_bus/issues/632
Things required to actually estabilish connections between indexers (such as the validator observer equivalent). 

Objective here is to create the facilities required to mechanically dispatch request between indexers. It was done of the same standard of the validator manager. Since indexers are volatile (but less so then writers) we used event hooks to induce change on the manager itself. There was "bonus" refactor which moved the event handlers to Network since they were just a coordination between state and network objects. There was a lifecycle violation since the events were hooked up on the state but the hook was using the network module to perform operations. In that context state listeners need to be removed before the handlers else it will lead to an uncaught exception (we were closing the network before removing the listeners).

This is intended to deliver this task: https://github.com/Trac-Systems/main_settlement_bus/issues/628


src/core/consensus/protocols/ConsensusV1Protocol.js
```diff
@@ -0,0 +1,61 @@
...
+class ConsensusV1Protocol {
+    #channel;
+    #session;
+    #router;
+    #pendingRequestService;
+    #publicKeyHex;
+
+    constructor(router, connection, pendingRequestService) {
+        this.#router = router;
+        this.#pendingRequestService = pendingRequestService;
```
b-garbacz, line 14: dito

src/core/consensus/protocols/ConsensusV1Protocol.js
```diff
@@ -0,0 +1,61 @@
+import Protomux from 'protomux';
+import c from 'compact-encoding';
+import { encodeConsensusMessage } from '../../../codecs/consensus/v1/consensusV1OperationCodec.js';
+
+class ConsensusV1Protocol {
+    #channel;
+    #session;
+    #router;
+    #pendingRequestService;
```
b-garbacz, line 9: IndexerPendingRequestService

src/core/consensus/protocols/ConsensusV1Protocol.js
```diff
@@ -0,0 +1,61 @@
...
+    #publicKeyHex;
+
+    constructor(router, connection, pendingRequestService) {
```
b-garbacz, line 12: dito

src/index.js
```diff
@@ -195,24 +190,13 @@ export class MainSettlementBus extends ReadyResource {
     }
 
     async #stateEventsListener() {
-        this.#state.on(CustomEventType.IS_INDEXER, (publicKey) => {
-            if (this.#isClosing) return;
-            this.#network.disconnectValidatorPeer(publicKey, 'peer promoted to indexer');
-        });
```
b-garbacz, line 201: Why are we deleting this?

src/index.js
```diff
@@ -34,7 +33,6 @@ export class MainSettlementBus extends ReadyResource {
     #wallet;
     #network;
     #state;
-    #isClosing = false;
```
b-garbacz, line 37: Why are we deleting this?

src/core/state/State.js
```diff
@@ -230,6 +238,19 @@ class State extends ReadyResource {
         return Object.values(this.#base.system.indexers);
     }
 
+    async isKnownIndexer(target) {
```
b-garbacz: 1. The name of this method is a bit confusing. How does it relate to "knowing" an indexer? To me, it looks more like it checks whether the targeted `remotePublicKey` currently belongs to an indexer.

2. I'm not sure if this method will be used anywhere else or if it should even be part of the `state.js` API. It could be used later in the validator handler, but I think it would be better if we checked whether the `targetAddress` matches the address instead. That way, the same API could be reused in both places.

What do you think about slightly changing the logic and comparing addresses instead of public keys?

src/core/state/State.js
```diff
@@ -210,6 +213,11 @@ class State extends ReadyResource {
         return !!adminEntry && this.#wallet?.address === adminEntry?.address && b4a.equals(adminEntry?.wk, this.writingKey)
     }
 
+    async isAdminPublicKey(publicKey) {
```
b-garbacz: In most of our API, we use `address` as the input argument. Let's keep that convention here as well.

src/core/state/State.js
```diff
@@ -210,6 +213,11 @@ class State extends ReadyResource {
...
+        const adminEntry = await this.getAdminEntry();
+        return !!adminEntry && b4a.equals(publicKey, tracCryptoApi.address.decodeSafe(adminEntry.address));
```
b-garbacz: Why are we decoding safe? If this is not part apply contract part then this should throw!

src/core/state/State.js
```diff
@@ -230,6 +238,19 @@ class State extends ReadyResource {
         return Object.values(this.#base.system.indexers);
     }
 
+    async isKnownIndexer(target) {
+        const entries = await this.getIndexersEntry();
+        for (const entry of entries) {
+            const address = await this.get(EntryType.WRITER_ADDRESS + b4a.toString(entry.key, 'hex'));
```
b-garbacz: Hmnn, why get? Do you want to get signed or unsigned?

src/config/config.js
```diff
@@ -107,6 +107,10 @@ export class Config {
         return this.#config.maxValidators
     }
 
+    get maxIndexers() {
```
b-garbacz: `maxIndexers` should not be the source of truth for indexer membership. The current indexer set should be derived from `base.system.indexers`; otherwise a static config value can
  diverge from Autobase state and cause valid indexers to be rejected. If this is meant only as a transport/resource limit, it should be named and enforced separately, and must not be
  lower than the current indexer set.

src/config/env.js
```diff
@@ -24,10 +24,13 @@ const configData = {
         pollInterval: 500, // Validator observer poll interval
         adminCacheTTL: 10_000, // Admin cache TTL ms
         validatorConnectionAttemptDelay: 5, // Delay between validator connection attempts (ms)
+        connectTimeoutMs: 5_000,
         bootstrapTimeout: 60_000, // time used (ms) to connect to new validators at bootstrap
         writersShortCacheTTL: 10_000, // short TTL during bootstrap
         writersLongCacheTTL: 120_000, // long TTL after bootstrap
         maxValidators: 50,
+        maxIndexers: 14,
```
b-garbacz: Dito

src/core/consensus/protocols/ConsensusRouter.js
```diff
@@ -11,13 +11,15 @@ class ConsensusRouterV1 {
     constructor(
         state,
         wallet,
-        config
+        config,
+        pendingRequestService
```
b-garbacz, line 15: IndexerPendingRequestService

src/core/consensus/protocols/ConsensusRouter.js
```diff
@@ -11,13 +11,15 @@ class ConsensusRouterV1 {
...
     ) {
         this.#config = config
-        this.#epochProofProposalHandler = new V1EpochProofProposalOperationHandler(
+        this.#epochProofProposalHandler = new ConsensusEpochProofProposalOperationHandler(
             state,
             wallet,
-            config
+            config,
+            pendingRequestService,
```
b-garbacz, line 22: dito

src/core/network/Network.js
```diff
@@ -114,24 +119,32 @@ class Network extends ReadyResource {
...
-        if (this.#swarm !== null) {
-            this.#swarm.destroy();
-        }
+        await this.#swarm.destroy();
     }
 
     setupNetworkListeners() {
+        this.#state.on(CustomEventType.IS_INDEXER, (publicKey) => {
+            this.addIndexerPeer(publicKey);
+            this.disconnectValidatorPeer(publicKey, 'peer promoted to indexer');
```
leonardostsouza, line 138: It would be better to disconnect Validator before adding as indexer, so we avoid the possibility of having the same node in 2 different pools, even if for some time.

src/core/network/Network.js
```diff
@@ -114,24 +119,32 @@ class Network extends ReadyResource {
...
+        });
+
+        this.#state.on(CustomEventType.IS_NON_INDEXER, (publicKey) => {
+            this.disconnectIndexerPeer(publicKey, 'peer demoted from indexer');
+        });
+
+        this.#state.on(CustomEventType.UNWRITABLE, (publicKey) => {
+            this.disconnectValidatorPeer(publicKey, 'peer became unwritable');
```
leonardostsouza, line 150: ALso, we set up a bunch of state listeners here in Network. Is this the best place to set them up?

If so, maybe it would be best to also clean them up in `cleanupNetworkListeners`. We setup these listeners but I didn't find where they are being cleaned up

src/config/config.js
```diff
@@ -227,6 +231,10 @@ export class Config {
         return this.#config.maxPendingRequestsInPendingRequestsService
     }
 
+    get maxPendingConnections() {
```
b-garbacz, line 234: I'm not sure if this should be in the config. In every config we have the same value: 50 connections. If this is the same for every config and we do not plan to change it, why split it
  into 3 separate cases? Wouldn't it be better to keep it as a `CONSTANT`, like it was before?

src/config/config.js
```diff
@@ -250,6 +258,10 @@ export class Config {
         return this.#config.bootstrapTimeout
     }
 
+    get connectTimeoutMs() {
```
b-garbacz, line 261: also this guy. Dito

src/config/env.js
```diff
@@ -77,10 +80,13 @@ const configData = {
         pollInterval: 500, // Validator observer poll interval
         adminCacheTTL: 3_600_000, // Admin cache TTL ms
         validatorConnectionAttemptDelay: 5, // Delay between validator connection attempts (ms)
+        connectTimeoutMs: 5_000,
```
b-garbacz, line 83: dito

src/config/env.js
```diff
@@ -77,10 +80,13 @@ const configData = {
...
         bootstrapTimeout: 120_000, // time used (ms) to connect to new validators at bootstrap
         writersShortCacheTTL: 2_000, // short TTL during bootstrap
         writersLongCacheTTL: 120_000, // long TTL after bootstrap
         maxValidators: 50,
+        maxIndexers: 14,
+        maxPendingConnections: 50,
```
b-garbacz, line 88: dito

src/config/env.js
```diff
@@ -130,10 +136,13 @@ const configData = {
         pollInterval: 500, // Validator observer poll interval
         adminCacheTTL: 60_000, // Admin cache TTL ms
         validatorConnectionAttemptDelay: 5, // Delay between validator connection attempts (ms)
+        connectTimeoutMs: 5_000,
```
b-garbacz, line 139: dito

src/config/env.js
```diff
@@ -130,10 +136,13 @@ const configData = {
...
         bootstrapTimeout: 30_000,  // time used (ms) to connect to new validators at bootstrap
         writersShortCacheTTL: 1_000, // short TTL during bootstrap
         writersLongCacheTTL: 2_000, // long TTL after bootstrap
         maxValidators: 10,
+        maxIndexers: 14,
+        maxPendingConnections: 50,
```
b-garbacz, line 144: dito

src/config/env.js
```diff
@@ -24,10 +24,13 @@ const configData = {
         pollInterval: 500, // Validator observer poll interval
         adminCacheTTL: 10_000, // Admin cache TTL ms
         validatorConnectionAttemptDelay: 5, // Delay between validator connection attempts (ms)
+        connectTimeoutMs: 5_000,
```
b-garbacz, line 27: dito

src/core/consensus/protocols/ConsensusMessages.js
```diff
@@ -1,22 +1,21 @@
-// Use this a factory to create the object attached to the swam connection
+import ConsensusRouterV1 from "./ConsensusRouter.js";
+import ConsensusV1Protocol from "./ConsensusV1Protocol.js";
+
 class ConsensusMessages {
-    // #config;
-    // #wallet;
-    // #state;
+    #consensusRouter;
+    #pendingRequestService;
```
b-garbacz, line 6: IndexerPendingRequestService

src/core/consensus/protocols/ConsensusMessages.js
```diff
@@ -1,22 +1,21 @@
...
-    //     state,
-    //     wallet,
-    //     config
-    // ) {
-    //     this.#config = config;
-    //     this.#wallet = wallet;
-    //     this.#state = state;
-    // }
+    constructor(state, wallet, config, pendingRequestService) {
+        this.#pendingRequestService = pendingRequestService;
```
b-garbacz, line 9: dito

src/core/consensus/protocols/ConsensusMessages.js
```diff
@@ -1,22 +1,21 @@
...
+        this.#consensusRouter = new ConsensusRouterV1(state, wallet, config, pendingRequestService);
```
b-garbacz, line 10: dito

src/core/consensus/protocols/ConsensusMessages.js
```diff
@@ -1,22 +1,21 @@
...
+    }
 
     async setupProtomuxMessages(connection) {
-        // The purpose of this is to grab a hyperswarm connection and enrich with an extra property that handle operations and carry metadata. We will need a specific class whose instnance will be created here.
-        connection.protocolSession;
+        connection.consensusProtocolSession = new ConsensusV1Protocol(
+            this.#consensusRouter,
+            connection,
+            this.#pendingRequestService
```
b-garbacz, line 17: dito

tests/unit/consensus/services/IndexerPendingRequestService.test.js
leonardostsouza, line 1: Please, add this file to `consensusModule.test.js`, otherwise these tests won't be run

tests/unit/consensus/consensusModule.test.js
```diff
@@ -5,6 +5,7 @@ async function runConsensusModuleTests() {
     await import('../consensus/ConsensusValidationSchema.test.js');
     await import('../consensus/V1EpochProofProposalRequest.test.js');
     await import('../consensus/V1EpochProofProposalApproval.test.js');
+    await import('./services/IndexerConnectionManager.test.js');
```
leonardostsouza, line 8: Please, add `IndexerPendingRequestService.test.js` here

src/core/consensus/services/IndexerPendingRequestService.js
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    getPendingRequest(id) {
+        const entry = this.#pendingRequests.get(id);
+        if (!entry) return null;
+        return entry;
+    }
+
+    // for now, we are resolving only resultCode, but we can extend it in the future if needed...
+    resolvePendingRequest(id, resultCode = ResultCode.OK) {
```
b-garbacz: ConsensusResultCode

src/core/consensus/services/IndexerPendingRequestService.js
```diff
@@ -0,0 +1,157 @@
...
+        return true;
+    }
+
+    close() {
+        for (const [id, entry] of this.#pendingRequests) {
+            clearTimeout(entry.timeoutId);
+            try {
+                entry.reject(
+                    new V1ProtocolError(
+                        ResultCode.UNEXPECTED_ERROR,
```
b-garbacz: dito

src/core/consensus/services/IndexerPendingRequestService.js
```diff
@@ -0,0 +1,157 @@
...
+        entry.timeoutId = null;
+        return true;
+    }
+
+    close() {
+        for (const [id, entry] of this.#pendingRequests) {
+            clearTimeout(entry.timeoutId);
+            try {
+                entry.reject(
+                    new V1ProtocolError(
```
b-garbacz: Use ConsensusVProtocolError

src/core/consensus/services/IndexerPendingRequestService.js
```diff
@@ -0,0 +1,157 @@
...
+
+        const promise = new Promise((resolve, reject) => {
+            entry.resolve = resolve;
+            entry.reject = reject;
+        });
+
+        entry.timeoutId = setTimeout(() => {
+            this.rejectPendingRequest(
+                id,
+                new IndexerPendingRequestServiceTimeoutError(
```
b-garbacz, line 72: I think `pendingRequestTimeout` should be different for the indexer layer. It should not use the same timeout as the validator/network pending request service, because other indexers may need more time to respond.

Also, this should probably be integrated with the state machine, so we clearly define how timeouts affect the proof proposal flow and quorum collection.

At the moment, this looks like a 1:1 copy paste from the regular pending request service.

src/core/network/Network.js
leonardotc: They are cleaned up by state. That is part of a clean up that was done when this was removed (from index)
```
        this.#state.on(CustomEventType.IS_INDEXER, (publicKey) => {
            if (this.#isClosing) return;
            this.#network.disconnectValidatorPeer(publicKey, 'peer promoted to indexer');
        });
```
The reson that `#isClosing` is there is because the network was being closed before the state so the listener would still exist while the network was still alive (resulting in an exception). Beyond that, imperative logic were leaking in a way that index would perform an orchestration between network and state. That chunk was moved so network performs its own operation when state changes (notified by event). Since we switched the order of close and moved the event cleaning to the "bus", state knows when to clean itself up, networks subscribes to what it needs and index just knows about which order the clean up should be done.
Overall, that is not good but I think its better then before and improving would require systems rework. We need: away to create singleton-like object in a clear lifecycle (and inject each component to those that need) and clear way to broadcast events.

src/core/network/Network.js
```diff
@@ -258,8 +279,13 @@ class Network extends ReadyResource {
                         publicKey,
                         new Error('Connection closed before response')
                     );
+                    this.#indexerPendingRequestService.rejectPendingRequestsForPeer(
```
b-garbacz, line 289: Okay. Just a reminder that the indexers sending approvals back to the requester will want to close the connection after some time. My intuition tells me this might not work well with the state machine. I'm not sure, though.

src/index.js
b-garbacz: Aaaa ok, did you just move it to the network.js. Right?

src/core/consensus/protocols/ConsensusRouter.js
```diff
@@ -29,7 +31,7 @@ class ConsensusRouterV1 {
         let decodedMessage;
 
         try {
-            decodedMessage = decodeV1networkOperation(incomingMessage)
+            decodedMessage = decodeConsensusMessage(incomingMessage)
         } catch (error) {
             this.#disconnect(connection, `Failed to decode incoming V1 message: ${error.message}`)
```
b-garbacz: This is not V1 protocol. If you will debug it you will be confused if this is V1 or Consensus protocol.

src/config/config.js
```diff
@@ -107,6 +107,11 @@ export class Config {
         return this.#config.maxValidators
     }
 
+    get maxIndexers() {
+        if (this.#isOverriden('maxIndexers')) return this.#options.maxIndexers
+        return this.#config.maxIndexers
```
leonardostsouza: This is a good question... For an epoch consensus point-of-view, do we want to have a "max connections" value?

src/core/consensus/protocols/ConsensusRouter.js
```diff
@@ -1,8 +1,8 @@
-import { decodeV1networkOperation } from '../../../codecs/network/v1/networkV1OperationCodec.js'
+import { decodeConsensusMessage } from '../../../codecs/consensus/v1/consensusV1OperationCodec.js'
 import b4a from 'b4a'
-import { NetworkOperationType, V1_PROTOCOL_PAYLOAD_MAX_SIZE } from '../../../utils/constants.js'
+import { ConsensusOperationType, CONSENSUS_MESSAGE_MAX_BYTE_SIZE } from '../../../utils/constants.js'
 import { publicKeyToAddress } from '../../../utils/helpers.js'
-import V1EpochProofProposalOperationHandler from './handlers/ConsesusEpochProofProposalOperationHandler.js'
+import ConsensusEpochProofProposalOperationHandler from '../v1/handlers/ConsesusEpochProofProposalOperationHandler.js'
 
 class ConsensusRouterV1 {
```
b-garbacz, line 7: If this class is 1:1 with `V1ProtocolRouter`, wouldn't it be better to create a base class that shares all these methods and place it in `main_settlement_bus/src/core/network/protocols/shared`?

  The only difference between them seems to be the `route` implementation. Also, `CONSENSUS_MESSAGE_MAX_BYTE_SIZE` and `V1_PROTOCOL_PAYLOAD_MAX_SIZE` have the same size.

  Maybe we should improve this?

src/core/consensus/protocols/ConsensusV1Protocol.js
```diff
@@ -0,0 +1,61 @@
...
+                    console.error(`ConsensusV1Protocol: unhandled router error: ${err.message}`);
+                    try { connection.end(); } catch {}
+                });
+            }
+        });
+    }
+
+    async send(message) {
+        const encodedMessage = encodeConsensusMessage(message);
+        const msgReplyPromise = this.#pendingRequestService.registerPendingRequest(this.#publicKeyHex, message);
```
b-garbacz, line 43: Will it be correlated with state machine???

b-garbacz, line 43: IndexerPendingRequestService

src/core/consensus/protocols/ConsensusV1Protocol.js
```diff
@@ -0,0 +1,61 @@
...
+        try {
+            this.#session.send(encodedMessage);
+        } catch (error) {
+            this.#pendingRequestService.rejectPendingRequest(message.session_id, error);
```
b-garbacz, line 47: dito

src/core/consensus/services/IndexerPendingRequestService.js
```diff
@@ -0,0 +1,157 @@
...
+
+        if (this.#pendingRequests.has(id)) {
+            throw new Error(`Pending request with ID ${id} from peer ${peerPubKeyHex} already exists.`);
+        }
+
+        const entry = {
+            id: id,
+            requestType: message.type,
+            requestedTo: peerPubKeyHex,
+            timeoutId: null,
```
b-garbacz, line 59: This would be good if we would like to store ProofProposal here.

src/core/network/Network.js
```diff
@@ -226,22 +239,30 @@ class Network extends ReadyResource {
...
+
             this.#swarm.on('connection', async (connection) => {
                 // Per-peer connection initialization:
                 // - attach Protomux (legacy + v1 channels/messages)
                 // - attach connection.protocolSession (used later by tryConnect / orchestrators to send messages)
-                await this.#networkMessages.setupProtomuxMessages(connection);
 
-                this.#consensusMessages.setupProtomuxMessages(connection);
+                // Adds indexers to consensus connection manager
+                if (await this.#state.isKnownIndexer(connection.remotePublicKey) && !await this.#state.isAdminPublicKey(connection.remotePublicKey)) {
```
b-garbacz: Dito, I referred to this in state.js

tests/unit/consensus/services/IndexerPendingRequestService.test.js
```diff
@@ -0,0 +1,55 @@
...
+
+test('registerPendingRequest stores the expected entry shape for proof proposals', async t => {
+    const service = new IndexerPendingRequestService(config);
+    const message = await buildProofProposal();
+
+    const promise = service.registerPendingRequest(validPeerA, message);
+    promise.catch(() => {});
+    t.teardown(() => service.close());
+
+    const entry = service.getPendingRequest(message.session_id);
```
b-garbacz, line 53: more tests?

tests/unit/consensus/services/IndexerConnectionManager.test.js
```diff
@@ -0,0 +1,150 @@
...
+const makeConnection = (publicKeyHex) => {
+    const emitter = new EventEmitter();
+    emitter.consensusProtocolSession = {
+        send: sinon.stub().resolves({ ok: true }),
+        sendAndForget: sinon.stub(),
+    };
+    emitter.remotePublicKey = b4a.from(publicKeyHex, 'hex');
+    return emitter;
+};
+
```
b-garbacz, line 17: More tests?

src/core/consensus/services/IndexerConnectionManager.js
```diff
@@ -0,0 +1,74 @@
...
+    constructor(config) {
+        this.#config = config;
+    }
+
+    add(publicKey, connection) {
+        const key = toHex(publicKey);
+        if (this.#indexers.has(key)) {
+            return false;
+        }
+        if (this.#indexers.size >= this.#config.maxIndexers) {
```
b-garbacz: dito. MaxIndexers is dynamic.

src/core/consensus/services/IndexerPendingRequestService.js
```diff
@@ -0,0 +1,157 @@
+import {ConsensusOperationType, ResultCode} from '../../../utils/constants.js';
+import {isHexString, publicKeyToAddress} from '../../../utils/helpers.js';
+import {V1ProtocolError} from "../../network/protocols/v1/V1ProtocolError.js";
+
+const PEER_PUBLIC_KEY_HEX_LENGTH = 64;
```
b-garbacz: Maybe we should have it in constants.js. Idk. But we have this same problem in validatorPendingRequestService

src/core/consensus/services/IndexerPendingRequestService.js
```diff
@@ -0,0 +1,157 @@
...
+        return this.#pendingRequests.has(id);
+    }
+
+    #validateRegisterInput(peerPubKeyHex, message) {
+        if (!isHexString(peerPubKeyHex) || peerPubKeyHex.length !== PEER_PUBLIC_KEY_HEX_LENGTH) {
+            throw new Error('Invalid peer public key. Expected 32-byte hex string.');
+        }
+
+        if (!message || typeof message !== 'object') {
+            throw new Error('Pending request message must be an object.');
```
b-garbacz: Will you be able to distinguish this error while debugging? What if ValidatorPendingRequestService will throw this error?

b-garbacz (commented):
Did you test it manually? Does it work?

src/index.js
leonardotc: That was there to prevent some execution of a module upon destruction during an event listener. We removed the network module while the listener was still hanging (we destroy network before state).

tests/unit/consensus/services/IndexerPendingRequestService.test.js
leonardotc: done.

tests/unit/consensus/consensusModule.test.js
leonardotc: done.

src/config/env.js
leonardotc: Because this was like this

```
this.#connectTimeoutMs = config.connectTimeoutMs || 5000;
this.#maxPendingConnections = config.maxPendingConnections || 50;
```
the data was not in the config so it was always going for the fallback while the tests were testing against the config.
```
    const config = {
        enableWallet: true,
        addressPrefix: 'trac',
        connectTimeoutMs: 1_000,
        maxPendingConnections: 10,
        maxValidators: 5,
        maxPeers: 5,
        maxParallel: 1,
        maxServerConnections: 5,
        maxClientConnections: 5,
        dhtBootstrap: [],
        channel: b4a.alloc(32, 1),
    };
```

src/core/network/Network.js
leonardotc: done.

src/config/config.js
leonardotc: I am not sure, this was done to use the same logic of maxValidators. We will probably have a max connections anyway due to swarm parameter peers parameter though.

leonardotc:
We tested manually but that was a while ago. The issue is to actually dispatch and handle requests we relied on a lot of working in progress code (some of which is still not delivered). So we tested and after rebasing with this: https://github.com/Trac-Systems/main_settlement_bus/pull/630 and in the same way a bunch of things changed in the dev branch afterwards (including renames that broke the branch and required fixing).

src/core/consensus/protocols/ConsensusRouter.js
leonardotc: Ok but this invalidates the previous comment of a super class. This place shouldnt know what kind of service is this. It just needs to know its something that matches pending request with response. I can change but it will be weird.

src/core/consensus/protocols/ConsensusRouter.js
leonardotc: Yes, it absolutely need to be like that.

src/core/consensus/services/IndexerPendingRequestService.js
leonardotc: done. I would advise to create a task to review constants. Many of them seem to have the same nature and sometimes its split.

src/core/consensus/services/IndexerPendingRequestService.js
leonardotc: done. Tbh, that look like ai code. I will just use safe navigation below and remove that. I think that was just copied from the other thing.

src/core/consensus/services/IndexerPendingRequestService.js
leonardotc: done.

src/core/consensus/services/IndexerPendingRequestService.js
leonardotc: done.

src/core/consensus/services/IndexerPendingRequestService.js
leonardotc: done.

src/config/config.js
lucasfernandes: Solved

src/core/state/State.js
lucasfernandes: Done. The method isKnownIndexer was renamed temporarily to isIndexerAddress.

src/core/state/State.js
lucasfernandes: Done. The method isAdminPublicKey was renamed to isAdminAddress.

src/core/state/State.js
lucasfernandes: Okay, agreed. Now the method does not decode anymore.

```
    async isAdminAddress(targetAddress) {
        const adminEntry = await this.getAdminEntry();
        return (adminEntry?.address === targetAddress);
    }
```

src/config/config.js
lucasfernandes: Okay, done.

src/config/env.js
lucasfernandes: Okay, done.

src/core/consensus/protocols/ConsensusRouter.js
lucasfernandes: Okay, done.

src/core/network/Network.js
lucasfernandes: Okay, done.

src/core/consensus/services/IndexerConnectionManager.js
lucasfernandes: Okay, done

src/core/consensus/services/IndexerPendingRequestService.js
lucasfernandes: Okay, done.

src/core/consensus/services/IndexerPendingRequestService.js
lucasfernandes: Done, created the indexerPendingRequestTimeout to be used there.

src/core/consensus/protocols/ConsensusRouter.js
lucasfernandes: This is expected to enter in the refactoring task #640

src/core/state/State.js
lucasfernandes: Ok, done

src/core/consensus/protocols/ConsensusRouter.js
b-garbacz: Didn't we want to rename `pendingRequestService` to `pendingRequestServiceInstance`? I think we discussed this during our last call with @leonardotc.

Alternatively, we could keep the instance name as `pendingRequestService` and rename the `PendingRequestService` class to `ValidatorPendingRequestService`. That way, `pendingRequestService` can remain a generic instance name. Probably this would be easier. What do you think about this?

src/core/consensus/protocols/ConsensusRouter.js
b-garbacz: Oki.

src/core/network/Network.js
```diff
@@ -35,8 +37,6 @@ class Network extends ReadyResource {
     #validatorMessageOrchestrator;
     #config;
     #pendingConnections;
-    #connectTimeoutMs;
-    #maxPendingConnections;
     #rateLimiter;
     #pendingRequestsService;
```
b-garbacz: Following up on my previous comment, we could rename this property to `validatorPendingRequestService`.

src/core/network/Network.js
```diff
@@ -61,15 +63,14 @@ class Network extends ReadyResource {
...
         this.#wallet = wallet
-        this.#connectTimeoutMs = config.connectTimeoutMs || 5000;
-        this.#maxPendingConnections = config.maxPendingConnections || 50;
         this.#pendingConnections = new Map();
         this.#transactionCommitService = new TransactionCommitService(this.#config);
         this.#transactionPoolService = new TransactionPoolService(state, wallet?.address, this.#transactionCommitService ,this.#config);
         this.#validatorObserverService = new ValidatorObserverService(this, state, wallet?.address, this.#config);
         this.#validatorConnectionManager = new ConnectionManager(this.#config);
         this.#validatorMessageOrchestrator = new MessageOrchestrator(this.#validatorConnectionManager, state, this.#config);
         this.#pendingRequestsService = new PendingRequestService(this.#config);
```
b-garbacz: and rename class name of `PendingRequestService` to `ValidatorPendingRequestService`

src/core/state/State.js
```diff
@@ -126,6 +129,10 @@ class State extends ReadyResource {
         return this.#base.isIndexer;
     }
 
+    async indexerCount() {
+        return await this.getIndexersEntry().length;
```
b-garbacz: Question: is this implementation covered by tests?

src/core/state/State.js
```diff
@@ -210,6 +217,11 @@ class State extends ReadyResource {
         return !!adminEntry && this.#wallet?.address === adminEntry?.address && b4a.equals(adminEntry?.wk, this.writingKey)
     }
 
+    async isAdminAddress(targetAddress) {
+        const adminEntry = await this.getAdminEntry();
+        return (adminEntry?.address === targetAddress);
```
b-garbacz, line 222: dito - tests

src/core/state/State.js
```diff
@@ -230,6 +242,15 @@ class State extends ReadyResource {
         return Object.values(this.#base.system.indexers);
     }
 
+    async isIndexerAddress(targetAddress) {
+        const entries = await this.getIndexersEntry();
+        for (const entry of entries) {
+            const address = await this.getSigned(EntryType.WRITER_ADDRESS + b4a.toString(entry.key, 'hex'));
```
b-garbacz, line 248: dito - tests

src/core/network/Network.js
lucasfernandes: Okay, renamed

src/core/network/Network.js
lucasfernandes: Okay, renamed.

src/core/state/State.js
lucasfernandes: Added tests

src/core/state/State.js
lucasfernandes: Added tests

src/core/state/State.js
lucasfernandes: Added tests

b-garbacz (approved):
I recommend to request review from copilot. This little buddy can notice interesting things. You can click it here:

Image

However, by my side LGTM.

PR #639, opened 2026-06-25
feat: create indexer observer service

lucasfernandes:
**Implement IndexerObserverService**

Adds `IndexerObserverService` to proactively ensure indexer nodes maintain connections to all other indexers, mirroring the pattern used by `ValidatorObserverService`.

Refactoring even driven structure into a plain algorithm on Network.js

### DIFF
--- a/src/config/env.js
+++ b/src/config/env.js
@@ -20,7 +20,6 @@ const configData = {
         channel: '1111trac1network1msb1testnet1111',
         dhtBootstrap: ['116.202.214.149:10001','157.180.12.214:10001','node1.hyperdht.org:49737','node2.hyperdht.org:49737','node3.hyperdht.org:49737'], // these are used to peer discovery
         derivationPath: address.TESNET_DERIVATION_PATH,
-        enableValidatorObserver: true,
         pollInterval: 500, // Validator observer poll interval
         adminCacheTTL: 10_000, // Admin cache TTL ms
         validatorConnectionAttemptDelay: 5, // Delay between validator connection attempts (ms)
@@ -74,7 +73,6 @@ const configData = {
         channel: '0000trac0network0msb0mainnet0000',
         dhtBootstrap: ['116.202.214.149:10001','157.180.12.214:10001','node1.hyperdht.org:49737','node2.hyperdht.org:49737','node3.hyperdht.org:49737'],
         derivationPath: address.MAINNET_DERIVATION_PATH,
-        enableValidatorObserver: true,
         pollInterval: 500, // Validator observer poll interval
         adminCacheTTL: 3_600_000, // Admin cache TTL ms
         validatorConnectionAttemptDelay: 5, // Delay between validator connection attempts (ms)
@@ -128,7 +126,6 @@ const configData = {
         channel: '12312313123123',
         dhtBootstrap: ['116.202.214.149:10001','157.180.12.214:10001','node1.hyperdht.org:49737','node2.hyperdht.org:49737','node3.hyperdht.org:49737'],
         derivationPath: address.MAINNET_DERIVATION_PATH,
-        enableValidatorObserver: true,
         pollInterval: 500, // Validator observer poll interval
         adminCacheTTL: 60_000, // Admin cache TTL ms
         validatorConnectionAttemptDelay: 5, // Delay between validator connection attempts (ms)
--- a/src/core/consensus/services/IndexerObserverService.js
+++ b/src/core/consensus/services/IndexerObserverService.js
@@ -1,0 +1,106 @@
+import SchedulableService from "../../../utils/scheduler/SchedulableService.js";
+import { Logger } from "../../../utils/logger.js";
+import { CONNECTION_STATUS } from "../../../utils/constants.js"
+import b4a from "b4a";
+import { bufferToAddress } from "../../../core/state/utils/address.js";
+import tracCryptoApi from "trac-crypto-api";
+import { publicKeyToAddress } from "../../../utils/helpers.js";
+
+
+class IndexerObserverService extends SchedulableService {
+    #network
+    #state
+    #config
+    #logger
+    #address
+
+    constructor(network, state, address, config) {
+        super();
+
+        this.#network = network;
+        this.#state = state;
+        this.#address = address;
+        this.#config = config;
+        this.#logger = new Logger(config);
+    }
+
+    #shouldRun() {
+        return !this.isInterrupted;
+    }
+
+    getScheduleInterval() {
+        return this.#config.pollInterval;
+    }
+
+    async start() {
+        super.start();
+        this.#logger.info("IndexerObserverService started");
+    }
+
+    async stop(waitForCurrent = true) {
+        const stopped = await super.stop(waitForCurrent);
+        if (!stopped) return;
+        this.#logger.info("IndexerObserverService stopped");
+    }
+
+    async #tryConnect(candidate) {
+        try {
+            const result = await this.#network.tryConnect(candidate.publicKeyHex, "indexer");
+            if (result && result !== CONNECTION_STATUS.IGNORED) {
+                this.#logger.info(`IndexerObserver: connected to indexer ${publicKeyToAddress(candidate.publicKeyHex, this.#config)} (${result})`);
+            }
+            return result;
+        } catch (err) {
+            this.#logger.error(`Indexer connection attempt failed: ${err.message}`);
+        }
+    }
+
+    async #getIndexersCandidates() {
+        const entries = await this.#state.getIndexersEntry();
+        const candidates = [];
+
+        for (const entry of entries) {
+            const writerKeyHex = b4a.toString(entry.key, 'hex');
+            const addressBuffer = await this.#state.getRegisteredWriterKey(writerKeyHex);
+            if (!addressBuffer) continue;
+
+            const addr = bufferToAddress(addressBuffer, this.#config.addressPrefix);
+            if (!addr || addr === this.#address) continue;
+
+            const publicKey = tracCryptoApi.address.decode(addr);
+            const publicKeyHex = b4a.toString(publicKey, 'hex');
+            candidates.push({ publicKey, publicKeyHex });
+        }
+
+        return candidates;
+    }
+
+    async #processCandidates(candidates) {
+        for (const candidate of candidates) {
+            if (!this.#shouldRun()) break;
+
+            const isConnected = this.#network.indexerConnectionManager.connected(candidate.publicKey);
+            const isPending = this.#network.isConnectionPending(candidate.publicKeyHex);
+
+            if (isConnected || isPending) continue;
+
+            await this.#tryConnect(candidate);
+        }
+    }
+
+    async worker(next) {
+        const interval = this.#config.pollInterval;
+        if (!this.#shouldRun()) return next(interval);
+
+        try {
+            const candidates = await this.#getIndexersCandidates();
+            if (candidates.length > 0) await this.#processCandidates(candidates);
+        } catch (err) {
+            this.#logger.error(`IndexerObserver worker error: ${err.message}`);
+        }
+
+        next(interval);
+    }
+}
+
+export default IndexerObserverService;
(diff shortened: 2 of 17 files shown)


src/config/env.js
```diff
@@ -21,6 +21,7 @@ const configData = {
         dhtBootstrap: ['116.202.214.149:10001','157.180.12.214:10001','node1.hyperdht.org:49737','node2.hyperdht.org:49737','node3.hyperdht.org:49737'], // these are used to peer discovery
         derivationPath: address.TESNET_DERIVATION_PATH,
         enableValidatorObserver: true,
+        enableIndexerObserver: true,
```
leonardostsouza: As most users are not indexers and will probably never mess with this file, I would only leave this on by default in `ENV.DEVELOPMENT`

src/config/env.js
```diff
@@ -77,6 +78,7 @@ const configData = {
         dhtBootstrap: ['116.202.214.149:10001','157.180.12.214:10001','node1.hyperdht.org:49737','node2.hyperdht.org:49737','node3.hyperdht.org:49737'],
         derivationPath: address.MAINNET_DERIVATION_PATH,
         enableValidatorObserver: true,
+        enableIndexerObserver: true,
```
leonardostsouza: As most users are not indexers and will probably never mess with this file, I would only leave this on by default in `ENV.DEVELOPMENT`

src/core/network/Network.js
```diff
@@ -400,36 +337,60 @@ class Network extends ReadyResource {
...
-                !this.#validatorPendingRequestService.isProbePending(connection.remotePublicKey.toString('hex'))
-            ) {
-                await this.#finalizeConnection(publicKey, type, connection);
-                return CONNECTION_STATUS.CONNECTED;
-            }
-        }
+        const connection = this.#swarm._allConnections.get(peerInfo.publicKey);
+        if (!connection) return CONNECTION_STATUS.PENDING;
+
+        const isConnectionReady = (connection.protocolSession && !this.#validatorPendingRequestService.isProbePending(connection.remotePublicKey.toString('hex'))) || connection.consensusProtocolSession
```
leonardostsouza, line 354: I'm unsure about this...  We trat a connection as "ready" if it has either a network `protocolSession` or a `consensusProtocolSession`. 

Maybe we should enforce `consensusProtocolSession`for indexers and `protocolSession`for validators.... or maybe this is unnecessary. WHat do you think?

src/config/env.js
lucasfernandes: we managed to kill this flag

src/config/env.js
lucasfernandes: we managed to kill this flag

src/core/network/Network.js
lucasfernandes: This will be solved in the upcoming pull request. It is part of much bigger refactor.

src/core/consensus/services/IndexerObserverService.js
```diff
@@ -0,0 +1,108 @@
...
+
+        for (const entry of entries) {
+            const writerKeyHex = b4a.toString(entry.key, 'hex');
+            const addressBuffer = await this.#state.getRegisteredWriterKey(writerKeyHex);
+            if (!addressBuffer) continue;
+
+            const addr = bufferToAddress(addressBuffer, this.#config.addressPrefix);
+            if (!addr || addr === this.#address || addr === adminEntry?.address) continue;
+
+            const publicKey = tracCryptoApi.address.decodeSafe(addr);
```
b-garbacz: We shouldn't use safe functions in non apply code. Check please `tracCryptoApi.address.decode`

src/core/consensus/services/IndexerObserverService.js
```diff
@@ -0,0 +1,108 @@
...
+        const adminEntry = await this.#state.getAdminEntry();
+        const candidates = [];
+
+        for (const entry of entries) {
+            const writerKeyHex = b4a.toString(entry.key, 'hex');
+            const addressBuffer = await this.#state.getRegisteredWriterKey(writerKeyHex);
+            if (!addressBuffer) continue;
+
+            const addr = bufferToAddress(addressBuffer, this.#config.addressPrefix);
+            if (!addr || addr === this.#address || addr === adminEntry?.address) continue;
```
b-garbacz: One of indexers is admin. Can't we process him?

src/core/consensus/services/IndexerObserverService.js
```diff
@@ -0,0 +1,108 @@
...
+        const stopped = await super.stop(waitForCurrent);
+        if (!stopped) return;
+        this.#logger.info("IndexerObserverService stopped");
+    }
+
+    async #tryConnect(candidate) {
+        try {
+            const result = await this.#network.tryConnect(candidate.publicKeyHex, "indexer");
+            if (result && result !== CONNECTION_STATUS.IGNORED) {
+                this.#logger.info(`IndexerObserver: connected to indexer ${candidate.publicKeyHex.slice(0, 8)} (${result})`);
```
b-garbacz: I recommend to display address instead of public key with `publicKeyToAddress` in helpers.js

src/core/consensus/services/IndexerObserverService.js
```diff
@@ -0,0 +1,108 @@
+import SchedulableService from "../../../utils/scheduler/SchedulableService.js";
+import { Logger } from "../../../utils/logger.js";
+import { CONNECTION_STATUS } from "../../../utils/constants.js"
+import b4a from "b4a";
+import { bufferToAddress } from "../../../core/state/utils/address.js";
+import tracCryptoApi from "trac-crypto-api";
+
+
+class IndexerObserverService extends SchedulableService {
```
b-garbacz, line 10: One general question: this class seems to work similarly to ValidatorObserverService, meaning it periodically tries to discover and connect to indexers.

Is this the intended behavior? I thought the design was that indexers should connect to each other only when VDF propagation is needed. In that model, the requester, for example the leader, would establish connections to other indexers in response to a specific event, such as completing the VDF computation.

With the current implementation, indexers remain continuously connected and keep trying to discover each other in the background.

TBD.

src/core/network/protocols/ProtocolSession.js
```diff
@@ -136,10 +136,7 @@ class ProtocolSession {
      * @returns {Boolean} True if health checks are supported in the preferred protocol, false otherwise.
      */
     isHealthCheckSupported() {
-        if (this.#preferredProtocol === null) {
-            throw new Error('ProtocolSession: preferred protocol not set. Call probe() first.');
-        }
-        return this.#preferredProtocol === this.#supportedProtocols.V1;
+        return this.isProbed() && this.#preferredProtocol === this.#supportedProtocols.V1;
```
b-garbacz, line 139: I’m not sure this change helps in this flow.

From what I see, `isHealthCheckSupported()` is called after we already try to `probe()` the connection. So if the session is still not probed at this point, that probably means something went wrong during protocol negotiation.

Returning `false` is safer because it avoids throwing, but it can also hide the issue and silently disable health checks for that validator. Maybe we should at least log this case, or keep the stricter behavior here.

@leonardostsouza What is your opinion about this???

src/core/network/Network.js
```diff
@@ -150,65 +150,24 @@ class Network extends ReadyResource {
             this.disconnectValidatorPeer(publicKey, 'peer became unwritable');
         });
 
-        this.on(EventType.VALIDATOR_CONNECTION_TIMEOUT, ({ publicKey, type, timeoutMs }) => {
-            this.#logger.debug(`Network Event: VALIDATOR_CONNECTION_TIMEOUT | PublicKey: ${publicKey} | Type: ${type} | TimeoutMs: ${timeoutMs}`);
```
b-garbacz, line 154: Could this race if protocol setup ever becomes asynchronous? `#promotePendingConnection()` assumes `connection.protocolSession` is already set, but protocol setup happens in a separate async `connection` listener.

@leonardostsouza

src/core/state/State.js
```diff
@@ -275,15 +275,6 @@ class State extends ReadyResource {
         return Object.values(this.#base.system.indexers);
     }
 
-    async isIndexerAddress(targetAddress) {
```
b-garbacz, line 369: It still has some references in tests

src/core/network/Network.js
```diff
@@ -111,6 +110,7 @@ class Network extends ReadyResource {
 
         if (this.#state.isIndexer() && !isAdmin) {
             this.#epochProofProposalService.start();
+            this.#indexerObserverService.start();
```
b-garbacz, line 111: admin can't use this service?

src/core/network/Network.js
```diff
@@ -150,65 +150,24 @@ class Network extends ReadyResource {
...
-                    this.#validatorHealthCheckService.stop(publicKey);
-                }
-            }
-
-        });
-
         this.#state.on(CustomEventType.IS_INDEXER, (bufferAddress) => {
             const address = tracCryptoApi.address.encode(this.#config.addressPrefix, bufferAddress);
             if (address === this.#wallet.address) {
+                this.#indexerObserverService.start();
```
b-garbacz, line 154: In line 107 admin can not use this service, but here it can. What if network have only one indexer and this is admin?

src/core/network/protocols/ProtocolSession.js
leonardostsouza: The `probe()` is the health check itself. We included the `isHealthCheckSupported()` because we could not guarantee that all validators would support v1 yet.

The indexers, however, represent a much more predictable group. We can just assume they support `probe()` already.

src/core/network/Network.js
leonardostsouza: It it becomes async, then yes. Otherwise, it doesn't seem to cause a race condition because `setupProtomuxMessages` functions are basically synchronous functions (although they are still marked as `async`).

Both the `protocolSession` and `consensusProtocolSession` are returned immediately.

src/core/consensus/services/IndexerObserverService.js
lucasfernandes: Okay, done.

src/core/consensus/services/IndexerObserverService.js
lucasfernandes: Okay, updated to process the admin

src/core/consensus/services/IndexerObserverService.js
lucasfernandes: Ok, displaying now the address

src/core/network/Network.js
lucasfernandes: Added a check to bypass admin

src/core/network/Network.js
lucasfernandes: In the current design we assumed that admin can't participate proposing epochs, only the genesis one. Should we allow it? if so, I can update the code accordingly.

src/core/state/State.js
lucasfernandes: Done, removed the remaining linked tests

src/core/network/protocols/ProtocolSession.js
lucasfernandes: so is this already addressed, right?

src/core/state/State.js
leonardotc: returned the thing

src/core/network/Network.js
leonardotc: Its kind of like that, yeah. Its on the refactor.

src/core/consensus/services/IndexerObserverService.js
leonardotc: We opened a task for this.

src/core/network/Network.js
lucasfernandes: Yes, I had to change this to sync because it was just an object creation. That solved the racing condition.

b-garbacz (changes requested):
Code looks good to me. However, didn’t we agree on the commit convention below?

```
<fix|docs|feature|chore>: <description>
```

b-garbacz (approved):

PR #660, opened 2026-06-30
refactor: move listeners from index.js to State.js

lucasfernandes:
This pr moves the listeners from index.js to State.js


b-garbacz (approved):
LGTM

Issue #647, opened 2026-06-26
feat: Implement new command (only for admin) and handler to process SET_GENESIS_EPOCH


b-garbacz:
# requisites 
https://github.com/Trac-Systems/main_settlement_bus/issues/643
https://github.com/Trac-Systems/main_settlement_bus/issues/644

# Acceptance criteria

- [ ] Implement new command with handler
- [ ] Cover handler with unit tests

PR #666, opened 2026-07-01
feat: Implement new command only for admin and handler to process set genesis epoch (#647)

b-garbacz:
# Description
This PR adds a new interactive CLI handler for genesis epoch initialization. The user can provide VDF parameters, confirm the operation, and broadcast the genesis epoch initialization operation to the MSB.

# Implemented

- Added interactive CLI flow for /init_genesis.
- Added validation and conversion for VDF parameters:
    - difficulty as uint32, 4 bytes,
    - discriminant bit size as uint16, 2 bytes.

- Added guards to prevent re-initialization when VDF parameters already exist.
- Added a new state key:
    - /parameters/vdf

- VDF parameters are stored as byte concatenation:
    - first 4 bytes: VDF difficulty,
    - next 2 bytes: VDF discriminant bit size.

- Added unit tests for the happy path and error cases.

 # Why
Genesis epoch initialization needs deterministic VDF parameters stored in ledger state. This allows consensus components to read the same parameters from state instead of relying only on local configuration.

### DIFF
--- a/src/codecs/apply/applyOperationCodec.js
+++ b/src/codecs/apply/applyOperationCodec.js
@@ -24,7 +24,7 @@ const normalizeDecodedApplyOperation = operation => {
     return operation;
 }
 
-const encodeApplyOperation = (payload) => {
+export const encodeApplyOperation = (payload) => {
     const error = Operation.verify(payload);
     if (error) throw new Error(error);
     return b4a.from(Operation.encode(payload).finish());
--- a/src/index.js
+++ b/src/index.js
@@ -1063,21 +1063,21 @@ export class MainSettlementBus extends ReadyResource {
 
     async handleEpochGenesisInitialization(params) {
         if (!this.#config.enableWallet) {
-            throw new Error("Can not initialize an admin - wallet is not enabled.");
+            throw new Error("Can not initialize genesis epoch - wallet is not enabled.");
         }
 
         const adminEntry = await this.#state.getAdminEntry();
 
         if (!adminEntry) {
             throw new Error(
-                "Can set epoch genesis operation - admin has been not initialized."
+                "Can not initialize genesis epoch - admin has not been initialized."
             );
         }
         if (!this.#wallet) {
-            throw new Error("Can not perform recovery - wallet is not initialized.");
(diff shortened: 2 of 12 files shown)


src/index.js
```diff
@@ -1054,6 +1060,46 @@ export class MainSettlementBus extends ReadyResource {
...
+        const txValidity = await this.#state.getIndexerSequenceState();
+        const payload = await applyStateMessageFactory(this.#wallet, this.#config)
+            .buildCompleteSetGenesisEpochMessage(
+                this.#wallet.address,
+                txValidity,
+                vdfDifficultyBuffer,
+                vdfDiscriminantSizeBuffer,
+            )
+        const encodedPayload = encodeApplyOperation(payload);
+        await this.#state.append(encodedPayload);
```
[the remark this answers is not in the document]
b-garbacz: This is not a scope of this task

tests/unit/state/stateModule.test.js
```diff
@@ -15,6 +15,7 @@ async function runStateTests() {
     await import('./apply/state.apply.test.js');
     if (!isBare()) {
         await import('./stateAdminValidation.test.js');
+        await import('./vdfParams.test.js');
```
leonardostsouza, line 18: I believe you forgot to append this file

src/codecs/apply/applyOperationCodec.js
```diff
@@ -24,7 +24,7 @@ const normalizeDecodedApplyOperation = operation => {
     return operation;
 }
 
-const encodeApplyOperation = (payload) => {
+export const encodeApplyOperation = (payload) => {
```
leonardostsouza, line 27: Why are we exporting this (and why we didn't export this before)?

b-garbacz: I didn't export it because it wasn't necessary when I implemented it. If it needs to be exported, then I'll change the signature.

tests/unit/state/stateModule.test.js
b-garbacz: Thx

leonardostsouza (approved):

PR #667, opened 2026-07-01
feat: extend builder class for apply operations to build set vdf para…

jusufsuljic:
Implemented builder support for the SET_VDF_PARAMS apply operation.

This PR wires SET_VDF_PARAMS into the apply message builder flow by:
  - exposing OperationType.SET_VDF_PARAMS
  - mapping the operation to the vpo payload key
  - adding complete-message builder support
  - adding buildCompleteSetVdfParamsMessage() to the message director
  - adding builder/director tests for the new operation

The VDF parameter setters now use the protocol-defined byte sizes:
  - df: 4 bytes
  - db: 2 bytes


b-garbacz (approved):
LGTM

PR #668, opened 2026-07-02
feat: extend statevalidationschema.js with a new validation for set v…

jusufsuljic:
Implemented schema validation for the SET_VDF_PARAMS apply operation.

This adds validation for the vpo payload, enforcing the expected field structure and byte sizes for tx, txv, df, db, in, and is. 

It also adds fixtures and unit tests covering happy path, strict schema behavior, required fields, invalid types, address length, and field buffer lengths.

NOTE: Due to the latest development, we figured that setting the discriminant is a destructive operation to perform outside of a hard-fork so that field is completely omitted in the proto, builder and schema validation with this PR.


b-garbacz (approved):
lgtm. Schema looks good, however fix the tests

PR #672, opened 2026-07-03
refactor: pendingrequestservice and connectionmanager

leonardotc:
This is quite a pickle. The entities that trigger tryConnect (the Observer) has two main responsibilities. "Force" connections and qualify connections. To explain: the first is to trigger swarm joinPeer to connect with an enitity while the second is to tell us that entity is indexer or validator based on the ledger data (the data they used to trigger that tryConnect. Because connection is an two-side thingy, there is the counterpart to that phenomena which is that you receive an incoming connection without triggering tryConnect on your side but rather receiving a connection request/incoming message. That goes through swarm manufacturing but doesnt have the Observer qualification. In general, the recommended solution is to attach a ledger verification to a theoretical factory method/function that returns a connection, that will qualify based on that ledger data ANYTIME that there is a connection (being the source the ObserverValidator tryConnect or not). The Observer responsibility would be focused on just estabilishing the forceful connection dynamics. Problem is that such a function exists, its exposed by DHT but swallowed by swarm.


lucasfernandes:
this pr is rebased with #739 and can be closed

PR #675, opened 2026-07-06
refactor: 274 refactor init to reuse encode logicor opposite in nodeentryjs

### LINKED ISSUE #274: Refactor init to reuse encode logic(or opposite) in nodeEntry.js
There is a lot of overlapping code between this function and `encode`. Can't we rewrite `init` by invoking `encode` with constant default values?

_Originally posted by @leonardostsouza in https://github.com/Trac-Systems/main_settlement_bus/pull/273#discussion_r2375949218_

# Description
Currently, init and encode duplicate a lot of logic when constructing a node entry buffer
# FR
- Accepts: `writingKey`, `role`, `balance`, `license`, `stakedBalance` (with defaults).  
- Validates inputs; returns empty buffer if invalid.  
- Must reuse `encode` to avoid code duplication.  
- Output: buffer in the same format as `encode`. 

Or oposite maybe we need to wrap init into the encode, or maybe function with another name... The programmer should decide what will be better.

lucasfernandes:
init() now delegates to encode() (removing the duplicated buffer-building logic), and re-validates the raw role via isNodeRoleValid before mapping it to isWhitelisted/isWriter/isIndexer flags.

### DIFF
--- a/src/core/state/utils/nodeEntry.js
+++ b/src/core/state/utils/nodeEntry.js
@@ -9,9 +9,9 @@ export const ZERO_BALANCE = b4a.alloc(BALANCE_BYTE_LENGTH);
 export const ZERO_LICENSE = b4a.alloc(LICENSE_BYTE_LENGTH);
 
 
-const isWhitelisted = role => [NodeRole.INDEXER, NodeRole.WHITELISTED, NodeRole.WRITER].includes(role);
-const isWriter = role => [NodeRole.INDEXER, NodeRole.WRITER].includes(role);
-const isIndexer = role => [NodeRole.INDEXER].includes(role);
+const roleIsWhitelisted = role => [NodeRole.INDEXER, NodeRole.WHITELISTED, NodeRole.WRITER].includes(role);
+const roleIsWriter = role => [NodeRole.INDEXER, NodeRole.WRITER].includes(role);
+const roleIsIndexer = role => [NodeRole.INDEXER].includes(role);
 
 /**
  * Initializes a new node entry with given writing key and role and the balance is set to zero.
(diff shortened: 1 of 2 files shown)


src/core/state/utils/nodeEntry.js
leonardostsouza: This is minor, but please update the JSDoc to reflect the argument name change (from `writingKey` to `wk`)

src/core/state/utils/nodeEntry.js
```diff
@@ -20,37 +24,15 @@ export const ZERO_LICENSE = b4a.alloc(LICENSE_BYTE_LENGTH);
...
-
-        return nodeEntry;
-    } catch (error) {
-        console.error('Error initializing node entry:', error);
+export function init(wk, role, balance = ZERO_BALANCE, license = ZERO_LICENSE, stakedBalance = ZERO_BALANCE) {
+    if (!isNodeRoleValid(role)) {
+        console.error('Invalid node role provided');
         return b4a.alloc(0);
     }
+
```
leonardostsouza: Also minor, but there are some whitespaces in this line. It would be good to auto-lint this file

tests/unit/state/utils/nodeEntry.test.js
leonardostsouza, line 1: The new tests are good, but since we are at it, it would be nice to include some tests for other invalid inputs too (`wk`, `balance`, `licence` and `stakedBalance`)

lucasfernandes: Okay, done.

src/core/state/utils/nodeEntry.js
lucasfernandes: Okay, done.

src/core/state/utils/nodeEntry.js
lucasfernandes: Okay, done.

leonardotc:
I dont think this will be used, I am closing to reduce noise.

PR #676, opened 2026-07-06
fix: safeWriteUInt32BE and other functions

b-garbacz:


src/core/consensus/v1/validators/V1EpochProofProposalRequest.js
```diff
@@ -54,9 +54,15 @@ class V1EpochProofProposalRequest extends V1BaseConsensusOperation {
      * Validates that the proof proposal VDF parameters hash matches local config.
      */
     async validateProofProposalVdfParametersHash(proofProposal) {
+        const vdfDifficulty = safeWriteUInt32BE(this.#config.vdfDifficulty, 0);
+        const vdfDiscriminantSizeBits = safeWriteUInt32BE(this.#config.vdfDiscriminantSizeBits, 0);
+        if (vdfDifficulty.length !== 4 || vdfDiscriminantSizeBits.length !== 4) {
```
leonardotc: These things should be validated either on the config (which would make sense only if they are overridable) and everywhere else should trust.

src/core/network/protocols/v1/validators/V1BaseOperation.js
```diff
@@ -21,6 +21,19 @@ class V1BaseOperation {
         this.#v1ValidationSchema = new V1ValidationSchema();
     }
 
+    #encodeResultCode(result) {
+        if (!Object.values(ResultCode).includes(result)) {
+            throw new V1ProtocolError(ResultCode.INVALID_PAYLOAD, `Invalid result code: ${String(result)}`);
+        }
+
+        const resultCode = safeWriteUInt32BE(result, 0);
+        if (resultCode.length !== 4) {
```
leonardotc: This validation doesnt makse sense here. You are already validating even more strictly just above. The premise here is: we shouldnt include a number that cant be expressed as 4 bytes in the enum.

leonardotc (changes requested):
There is another thing which is a bit strange. I understand that there are places in the code where we return `buf.alloc(0)` as an invalid argument but beware of the consequences. That is a wrapped buffer array which js will not consider to be `falsy` (these guys https://developer.mozilla.org/en-US/docs/Glossary/Falsy). Furthermore in js, the comparison `new Uint8Array() === new Uint8Array()` will also return `false` since we are comparing two object pointers. Consider simply using something like null since doing a `!result` is clearer and safer then doing a `result.length !== 0`.

leonardotc (approved):

PR #678, opened 2026-07-08
feat: added command and handler for set vdf params

### LINKED ISSUE #670: feat: implement new ADMIN command and handler to process SET_VDF_PARAMS

jusufsuljic:
Adds the admin-only /set_vdf_params CLI flow for updating VDF difficulty after genesis VDF params already exist.

The command collects a new difficulty, asks for confirmation, validates admin/wallet/state requirements, builds a SET_VDF_PARAMS operation, encodes it, and appends it to state. It intentionally updates only difficulty and leaves discriminant unchanged.

Command and handler along with their tests are implemented.

### DIFF
--- a/cli/commandHandler.js
+++ b/cli/commandHandler.js
@@ -38,7 +38,8 @@ export const COMMANDS = {
     GET_TXS_HASHES: "/get_txs_hashes",
     GET_TX_DETAILS: "/get_tx_details",
     GET_EXTENDED_TX_DETAILS: "/get_extended_tx_details",
-    EPOCH_GENESIS_INITIALIZATION: "/init_genesis"
+    EPOCH_GENESIS_INITIALIZATION: "/init_genesis",
+    SET_VDF_PARAMS: "/set_vdf_params"
 };
 
 export class CommandHandler {
@@ -48,6 +49,7 @@ export class CommandHandler {
     #handlers;
     #pendingConfirmation = null;
     #pendingGenesisInitialization = null;
+    #pendingSetVdfParams = null;
 
     constructor({ config, msb, handleClose, wallet }) {
         this.#msb = msb;
@@ -63,6 +65,10 @@ export class CommandHandler {
 
         if (this.#pendingGenesisInitialization !== null) {
             return this.#handlePendingGenesisInitialization(input);
+        }
+
+        if (this.#pendingSetVdfParams !== null) {
+            return this.#handlePendingSetVdfParams(input);
         }
 
         const [command, ...parts] = input.split(" ");
@@ -230,6 +236,10 @@ export class CommandHandler {
             {
                 evaluate: ({ input }) => input.startsWith(COMMANDS.EPOCH_GENESIS_INITIALIZATION),
                 process: async () => this.#queueEpochGenesisInitialization()
+            },
+            {
+                evaluate: ({ input }) => input.startsWith(COMMANDS.SET_VDF_PARAMS),
+                process: async () => this.#queueSetVdfParams()
             }
         ];
     }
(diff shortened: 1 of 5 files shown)


cli/commandHandler.js
```diff
@@ -65,6 +67,10 @@ export class CommandHandler {
             return this.#handlePendingGenesisInitialization(input);
         }
 
+        if (this.#pendingSetVdfParams !== null) {
```
leonardotc, line 70: This is the kind of thing that started wrong and kept on giving.  This shouldnt have "exceptional flows". It should be described as a pair of evaluation and execution. In the worst case scenario, this should be:
```
            {
                evaluate: () => this.#pendingSetVdfParams !== null,
                process: async (input) => {
                    this.#handlePendingSetVdfParams(input)
                }
            },
```
Is there any particular reason we keeping adding exceptions to this flow?

src/index.js
```diff
@@ -292,6 +292,7 @@ export class MainSettlementBus extends ReadyResource {
             console.log("- /remove_indexer <address>: Change a role of the selected indexer node to default role. Charges a fee.");
             console.log("- /ban_writer <address>: Demote a whitelisted writer to default role and remove it from the whitelist. Charges a fee.");
             console.log("- /init_genesis: Initialize genesis epoch.");
+            console.log("- /set_vdf_params: Update VDF difficulty.");
```
leonardotc, line 295: This is an admin operation, right? Should it show here?

src/index.js
```diff
@@ -1110,6 +1111,46 @@ export class MainSettlementBus extends ReadyResource {
...
+        if (!this.#config.enableWallet) {
+            throw new Error("Can not set VDF params - wallet is not enabled.");
+        }
+
+        const adminEntry = await this.#state.getAdminEntry();
+
+        if (!adminEntry) {
+            throw new Error("Can not set VDF params - admin has not been initialized.");
+        }
+        if (!this.#wallet) {
```
leonardotc, line 1125: If the wallet is enabled, we should rely on this to be true. If it needs to be verified its probably somewhere else.

cli/commandHandler.js
```diff
@@ -48,6 +49,7 @@ export class CommandHandler {
     #handlers;
     #pendingConfirmation = null;
     #pendingGenesisInitialization = null;
+    #pendingSetVdfParams = null;
```
leonardotc, line 52: How long until we have two hundred instance variables that represents exceptional flows? I imagine this just means the current dynamics dont satisfy what you guys need. Maybe what you need is a shared context between command line calls (and that context would be handled, cleared and kept in between operations). The evaluate could reason about the context as much as it does about the command itself. Could even mark them as final etc etc. I think the third instance here just means the whole solution needs to be improved.

src/index.js
jusufsuljic: This is encased in an existing "adminMode" clause. I just fit it in there.

cli/commandHandler.js
jusufsuljic: I understand the point and I agree, it looks like it started wrong. We should definitely bring this out as an semi-urgent refactor chore that will not affect the core mechanism but only the UI/UX portion, meaning it can be done as an independent improvement.

Or if it is going to benefit us right now, I can do something as a part of this PR.

@leonardotc @b-garbacz @lucasfernandes @leonardostsouza

cli/commandHandler.js
jusufsuljic: Same as [this](https://github.com/Trac-Systems/main_settlement_bus/pull/678#discussion_r3543680709)

src/index.js
jusufsuljic: Agreed, we can go with a private wallet getter helper function that will ensure that we have the wallet before we get into the handlers and clean up the checks.

However, this is a repeating consistency/runtime check that is now bad, because it is repeated obviously, but imho falls out of the scope of this PR.

Like with the commandHandler situation, I would raise this as a semi-high priority chore.

b-garbacz (approved):

PR #682, opened 2026-07-10
refactor: implement shared protocol result codes and error handling

leonardostsouza:
## Description
The objective of this task is to unify consensus and network result codes in a single protobuf file

## Motivation
Epoch Consensus and Network were sharing a lot similar errors, but the error codes were confusing and often were not matching. This PR aims to solve this problem, avoiding result code name duplication and mismatch between similar errors in Network and epoch Consensus

## Scope
- This PR aims to unify network and epoch consensus result codes in a single file
- It also aims to expand available epoch consensus result codes, which were previously limited in scope. The objective is to have more specific results to return in case of errors. 
- Some of the result codes are still unused, but were added as placeholders. They are necessary for future validator/apply operation handler implementations.


src/utils/constants.js
```diff
@@ -62,77 +59,96 @@ export const ConsensusProtocolVersion = Object.freeze({
...
+    PENDING_REQUEST_MISSING_TX_DATA: SharedResultCode.RESULT_CODE_PENDING_REQUEST_MISSING_TX_DATA,
+    PROOF_PAYLOAD_MISMATCH: SharedResultCode.RESULT_CODE_PROOF_PAYLOAD_MISMATCH,
+    VALIDATOR_WRITER_KEY_NOT_REGISTERED: SharedResultCode.RESULT_CODE_VALIDATOR_WRITER_KEY_NOT_REGISTERED,
+    VALIDATOR_ADDRESS_MISMATCH: SharedResultCode.RESULT_CODE_VALIDATOR_ADDRESS_MISMATCH,
+    VALIDATOR_NODE_ENTRY_NOT_FOUND: SharedResultCode.RESULT_CODE_VALIDATOR_NODE_ENTRY_NOT_FOUND,
+    VALIDATOR_NODE_NOT_WRITER: SharedResultCode.RESULT_CODE_VALIDATOR_NODE_NOT_WRITER,
+    VALIDATOR_WRITER_KEY_MISMATCH: SharedResultCode.RESULT_CODE_VALIDATOR_WRITER_KEY_MISMATCH,
+    VALIDATOR_TX_OBJECT_INVALID: SharedResultCode.RESULT_CODE_VALIDATOR_TX_OBJECT_INVALID,
+    VALIDATOR_VA_MISSING: SharedResultCode.RESULT_CODE_VALIDATOR_VA_MISSING,
+    TX_INVALID_PAYLOAD: SharedResultCode.RESULT_CODE_TX_INVALID_PAYLOAD
```
b-garbacz, line 149: What do you think to create a new task to delete this constant from constants.js and just port it?

proto/common/v1/enums/result_code.proto
```diff
@@ -81,4 +87,25 @@ enum ResultCode {
   RESULT_CODE_VALIDATOR_TX_OBJECT_INVALID = 59;
   RESULT_CODE_VALIDATOR_VA_MISSING = 60;
   RESULT_CODE_TX_INVALID_PAYLOAD = 61;
+
+  // 100-199: [SCOPE: EPOCH_CONSENSUS_V1]
+  // 100-114: epoch consensus proof proposal validation outcomes.
+  RESULT_CODE_CONSENSUS_SCHEMA_VALIDATION_FAILED = 100;
```
b-garbacz, line 97: `RESULT_CODE_CONSENSUS_SCHEMA_VALIDATION_FAILED` is used in this same scenario `RESULT_CODE_INVALID_PAYLOAD` for schema validation. Any idea for this?

src/utils/constants.js
b-garbacz: Maybe it was bad idea in past to copy all of them

b-garbacz (commented):
Regarding the pre-check, this looks like a good direction to me.

src/utils/constants.js
leonardostsouza: The core idea here is that I was able to separate contexts from the shared proto file using these constants in contants.js.

This way, `ConsensusResultCode` does not expose codes exclusive to `NetworkResultCode` and vice versa even if they use the same proto file

proto/common/v1/enums/result_code.proto
leonardostsouza: "Invalid payload" is for when we would expect a payload for consensus/v1 and we would receive something that was not even identifiable (e.g.: missing type field)

"Schema Validation Failed" is for when we could identify the payload type, meaning all necessary fields are present, but the data doesn't fulfil the requirements (e.g.: some field in the payload has invalid length). In this case, we usually return a more specific error, but I wanted to have a "fallback" error to return in case of something unexpected

b-garbacz: V1_PROTOCOL shouldn't have then `RESULT_CODE_VALIDATOR_V1_SCHEMA_VALIDATION_FAILED`?

scripts/docs/check-v1-result-code-policies.js
```diff
@@ -84,26 +84,18 @@ function compareConstantsWithGeneratedProto() {
     const errors = [];
     const protoRows = buildGeneratedProtoResultCodes();
     const protoByName = new Map(protoRows.map((row) => [row.resultCodeName, row.value]));
-    const constantsByName = new Map(Object.entries(ResultCode));
+    const constantsByName = new Map(Object.entries(NetworkResultCode));
```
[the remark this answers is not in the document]
leonardostsouza: This is intentional

proto/consensus/v1/messages/proof_proposal_response.proto
```diff
@@ -2,13 +2,13 @@ syntax = "proto3";
...
-import "consensus/v1/enums/result_code.proto";
+import "common/v1/enums/result_code.proto";
 import "consensus/v1/messages/proof_proposal_approval.proto";
 
 // response sig = SIG(proof_proposal + approval_sig + result)
 
 message ProofProposalResponse {
-  ResultCode result = 1;
+  .common.v1.ResultCode result = 1;
   ProofProposalApproval approval = 2; // set only when result === OK
```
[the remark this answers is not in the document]
leonardostsouza: out of scope

src/utils/constants.js
b-garbacz: Yes but this is possible to reduce size of this...

docs/v1_result_code_policies.md
```diff
@@ -1,6 +1,8 @@
-# V1 Result Code Policies
+# Network V1 Result Code Policies
 
-This table shows, for each V1 `ResultCode`, the boolean policy flags that determine whether the sender rotates the validator and whether the validator closes the connection after sending the response.
+This table shows, for each Network V1 `NetworkResultCode`, the boolean policy flags that determine whether the sender rotates the validator and whether the validator closes the connection after sending the response.
+
+Network and consensus result codes are registered in the shared protobuf enum `common.v1.ResultCode`; this document intentionally covers only the Network V1 subset used by network connection policies.
 
 `true` in `Sender` means the sender rotates the validator for that result code. `true` in `Validator` means the validator closes the connection after sending the response. `false` means the corresponding action is not executed for that result code. These columns should be interpreted as boolean flags that control execution for each code, not as descriptive status fields.
```
b-garbacz, line 7: Let's create a new task to create a new documentation for consensus result codes, that does not need to be complex like this. However, it would be good, if script checker will validate them too

b-garbacz (approved):
LGTM

PR #684, opened 2026-07-10
feat: implement missing checks in validation module for internal consensus messages

### LINKED ISSUE #679: VDF parameters must be not stored in the ledger/ config file.
Right now, VDF parameters are stored directly in state. This works for the current Proof of Time/VDF flow, but it can become technical debt.

The current Autobase consensus and VDF based Proof of Time are temporary. We should avoid making VDF specific data a permanent part of the state model.

## Problem

The current state model is tied to one specific consensus mechanism. If we change the consensus implementation later, VDF state entries may make migration harder.

We also need to support different sync modes. Some nodes may rebuild state from history, while others may use fastForward sync. The future solution should work in both cases.

## Goal

Design a future proof way to manage consensus parameters.

The solution should use a generic, verifiable parameter model, for example based on a Merkle tree like.

Signed state should stay the source of truth for verification, but it should not need to store all consensus parameter details directly.

## High-Level Requirements

- Decouple persistent state from VDF parameters.
- Support future consensus mechanisms.
- Allow consensus parameters to be verified.
- Work with both history replay and fastForward sync.
- Avoid direct mutation of old consensus parameters.
- Keep the design generic and easy to extend.

## Notes

This issue is intentionally high-level. The exact data model and implementation details should be designed separately.

b-garbacz:
## Summary

- Added missing proof proposal validations for epoch, previous epoch hash, and indexer membership.
- Sourced VDF parameters from signed state to keep consensus nodes consistent.
- Added a shared error boundary to normalize unexpected failures into protocol errors.
- Made missing epoch and VDF state fail explicitly instead of returning null.
- Removed obsolete VDF parameters from local configuration.
- Updated protocol documentation and expanded unit test coverage.

### DIFF
--- a/src/core/consensus/v1/validators/V1BaseConsensusOperation.js
+++ b/src/core/consensus/v1/validators/V1BaseConsensusOperation.js
@@ -52,14 +52,17 @@ class V1BaseConsensusOperation {
                 throw error;
             }
 
-            const reason = error instanceof Error
-                ? error.message
-                : typeof error === 'string' ? error : String(error);
+            let validationError = error;
+            if (!(validationError instanceof Error)) {
+                validationError = new TypeError('Validation threw a non-Error value.');
+                validationError.cause = error;
+            }
+
             const protocolError = new V1ConsensusProtocolError(
                 ConsensusResultCode.UNEXPECTED_ERROR,
-                reason
+                validationError.message
             );
-            protocolError.cause = error;
+            protocolError.cause = validationError;
             throw protocolError;
         }
     }
--- a/src/core/consensus/v1/validators/V1BaseConsensusOperation.js
+++ b/src/core/consensus/v1/validators/V1BaseConsensusOperation.js
@@ -39,6 +39,7 @@ class V1BaseConsensusOperation {
      * Expected protocol errors keep their result code and identity. Unexpected
      * errors from state, codecs, crypto APIs, or malformed runtime values are
      * exposed as a stable protocol error while remaining available as `cause`.
+     * Validation operations are expected to throw `Error` instances.
      *
      * @param {function(): (Promise<*>|*)} validation Validation operation to execute.
      * @returns {Promise<*>} Result returned by the validation operation.
@@ -52,17 +53,11 @@ class V1BaseConsensusOperation {
                 throw error;
             }
 
-            let validationError = error;
-            if (!(validationError instanceof Error)) {
-                validationError = new TypeError('Validation threw a non-Error value.');
-                validationError.cause = error;
-            }
-
             const protocolError = new V1ConsensusProtocolError(
                 ConsensusResultCode.UNEXPECTED_ERROR,
-                validationError.message
+                error.message
             );
-            protocolError.cause = validationError;
+            protocolError.cause = error;
             throw protocolError;
         }
     }
--- a/src/core/consensus/v1/validators/V1BaseConsensusOperation.js
+++ b/src/core/consensus/v1/validators/V1BaseConsensusOperation.js
@@ -13,17 +13,66 @@ import b4a from "b4a";
 
 class V1BaseConsensusOperation {
     #consensusValidationSchema;
-    #config;
+    _config;
+    _state;
     /**
-     * Creates the base consensus validator with schema validation support.
+     * Creates the base validator shared by consensus v1 operations.
+     *
+     * Initializes operation schemas and stores the shared dependencies used by
+     * address, signature, and state-backed validation.
+     *
+     * @param {Config} config Application configuration. The base validator uses
+     * `addressLength` for schemas and `addressPrefix` for address conversion.
+     * @param {State} state Ledger state. It must expose `isIndexerAddress(address)`
+     * for indexer membership validation.
+     * @throws {Error} When consensus schemas cannot be initialized from the configuration.
      */
-    constructor(config) {
+    constructor(config, state) {
         this.#consensusValidationSchema = new ConsensusValidationSchema(config);
-        this.#config = config;
+        this._config = config;
+        this._state = state;
     }
 
     /**
-     * Validates the consensus payload shape for its declared operation type.
+     * Runs a validator behind the consensus protocol error boundary.
+     *
+     * Expected protocol errors keep their result code and identity. Unexpected
+     * errors from state, codecs, crypto APIs, or malformed runtime values are
+     * exposed as a stable protocol error while remaining available as `cause`.
+     *
+     * @param {function(): (Promise<*>|*)} validation Validation operation to execute.
+     * @returns {Promise<*>} Result returned by the validation operation.
+     * @throws {V1ConsensusProtocolError} Re-throws protocol errors or wraps unexpected failures.
+     */
+    async validateAsProtocolError(validation) {
+        try {
+            return await validation();
+        } catch (error) {
+            if (error instanceof V1ConsensusProtocolError) {
+                throw error;
+            }
+
+            const reason = error instanceof Error
+                ? error.message
+                : typeof error === 'string' ? error : '';
+            const protocolError = new V1ConsensusProtocolError(
+                ConsensusResultCode.UNEXPECTED_ERROR,
+                reason
+            );
+            protocolError.cause = error;
+            throw protocolError;
+        }
+    }
+
+    /**
+     * Validates the payload against the schema selected by its operation type.
+     *
+     * Checks that the operation type is present, supported, and that all fields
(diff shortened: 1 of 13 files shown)


src/index.js
```diff
@@ -1081,7 +1082,9 @@ export class MainSettlementBus extends ReadyResource {
             throw new Error("Can not initialize genesis epoch - you are not the admin.");
         }
 
-        const existingVDFParams = await this.#state.getSignedVDFParams();
+        // Read the raw entry because missing VDF params are expected before genesis,
+        // while getSignedVDFParams() intentionally throws when they are not initialized.
```
leonardostsouza: This is not a good behaviour. A "getXYZ" function is usually a query. so it would be reasonable for it to just return `null` when the entry does not exist. 

It you need this kind of function to throw, you could create something like:

```js
requireXYZ(arg) {
    const res = getXYZ(arg)
    if (_isNil.(res)) throw err;
    return res;
}
```

src/core/state/State.js
```diff
@@ -177,38 +178,65 @@ class State extends ReadyResource {
...
+            throw new Error(
+                'Cannot read epoch: epoch id is required.'
+            );
+        }
 
         const epochId = typeof count === 'bigint' ? count : BigInt(count);
-        return await this.getSigned(EntryType.EPOCH + epochId.toString());
+        const epochHash = await this.getSigned(EntryType.EPOCH + epochId.toString());
+        if (_.isNil(epochHash)) {
+            throw new Error(
```
leonardostsouza, line 234: Check https://github.com/Trac-Systems/main_settlement_bus/pull/684#discussion_r3569742359

src/core/state/State.js
```diff
@@ -177,38 +178,65 @@ class State extends ReadyResource {
...
      *
-     * @returns {Promise<bigint|null>} Current epoch id, or null when epoch state is not initialized.
+     * @returns {Promise<bigint>} Current epoch id.
+     * @throws {Error} When genesis epoch state has not been initialized.
      */
     async getCurrentEpoch() {
         const currentEpoch = await this.getSigned(EntryType.EPOCH_CURRENT);
-        if (currentEpoch === null) return null;
+        if (_.isNil(currentEpoch)) {
+            throw new Error(
```
leonardostsouza, line 198: Check https://github.com/Trac-Systems/main_settlement_bus/pull/684#discussion_r3569742359

src/core/state/State.js
```diff
@@ -177,38 +178,65 @@ class State extends ReadyResource {
...
-        if (epochHash === null || epochHash === undefined) return null;
+        if (_.isNil(epochHash)) {
+            throw new Error('Cannot read epoch proof: epoch hash is required.');
+        }
 
         const epochHashString = b4a.isBuffer(epochHash) ? epochHash.toString('hex') : epochHash.toString();
-        return await this.getSigned(EntryType.EPOCH_HASH + epochHashString);
+        const epochProof = await this.getSigned(EntryType.EPOCH_HASH + epochHashString);
+        if (_.isNil(epochProof)) {
+            throw new Error(
```
leonardostsouza, line 268: Check https://github.com/Trac-Systems/main_settlement_bus/pull/684#discussion_r3569742359

src/core/state/State.js
```diff
@@ -546,7 +574,9 @@ class State extends ReadyResource {
 
     async getSignedVDFParams() {
         const vdfParamsBuffer = await this.getSigned(EntryType.VDF_PARAMS);
-        if (!vdfParamsBuffer) return null;
+        if (!vdfParamsBuffer) {
+            throw new Error("VDF parameters are not initialized.");
```
leonardostsouza: Check https://github.com/Trac-Systems/main_settlement_bus/pull/684#discussion_r3569742359

src/index.js
leonardostsouza: Just to be clear, this is basically a naming convention issue

tests/unit/consensus/V1EpochProofProposalApproval.test.js
```diff
@@ -20,6 +21,9 @@ import {createMessage, uint32ToBuffer} from '../../../src/utils/buffer.js';
 const previousEpochRecordHash = b4a.alloc(32, 1);
 const vdfParametersHash = b4a.alloc(32, 2);
 const vdfProof = b4a.alloc(VDF_BLOB_PROOF_SIZE, 3);
+const state = {
+    isIndexerAddress: async () => true
```
leonardostsouza, line 25: Please, include a test case for when `isIndexerAddress` returns `false`

tests/unit/consensus/V1EpochProofProposalRequest.test.js
```diff
@@ -25,71 +30,116 @@ import {config} from '../../helpers/config.js';
...
+    vdfDiscriminantSize = TEST_VDF_PARAMS.vdfDiscriminantSize
+} = {}) {
+    return {
+        getCurrentEpoch: async () => currentEpoch,
+        getEpoch: async () => currentEpochHash,
+        getSignedVDFParams: async () => ({
+            vdfDifficulty,
+            vdfDiscriminantSize
+        }),
+        isIndexerAddress: async () => true
```
leonardostsouza: Please, include a test case for when `isIndexerAddress` returns `false`

src/core/consensus/v1/validators/V1BaseConsensusOperation.js
```diff
@@ -203,15 +287,29 @@ class V1BaseConsensusOperation {
...
+     * @param {Buffer} remotePublicKey Public key received from the peer connection.
+     * @returns {Promise<void>}
+     * @throws {V1ConsensusProtocolError} When the remote address is not an indexer.
+     */
+    async validateAddressIsIndexer(remotePublicKey) {
         // TODO: When we replace UNEXPECTED_ERROR with INVALID_ADDRESS_ASSERTION,
         // we should handle this specific error to not only drop the connection but also blacklist the specific node.
         // Such an error would mean that someone is trying to impersonate an indexer.
+        const address = tracCryptoApi.address.encode(this._config.addressPrefix, remotePublicKey);
+        const isIndexer  =  await this._state.isIndexerAddress(address);
```
leonardostsouza: Looks like there are double spaces here. Try auto-identing this code

src/core/consensus/v1/validators/V1BaseConsensusOperation.js
```diff
@@ -13,17 +13,66 @@ import b4a from "b4a";
...
+        try {
+            return await validation();
+        } catch (error) {
+            if (error instanceof V1ConsensusProtocolError) {
+                throw error;
+            }
+
+            const reason = error instanceof Error
+                ? error.message
+                : typeof error === 'string' ? error : String(error);
```
leonardostsouza: If the `error` is an object, doing `String(error)` will return `[object Object]`, so we may lose some context here. It would be better to enforce an error type coming from `validation()`

src/core/consensus/v1/validators/V1EpochProofProposalRequest.js
```diff
@@ -7,83 +7,132 @@ import b4a from "b4a";
...
+                connection.remotePublicKey
+            );
+            await this.validateSignature(payload, connection.remotePublicKey);
+            await this.validateAddressIsIndexer(connection.remotePublicKey);
+            const currentEpoch = await this._state.requireCurrentEpoch();
+            this.validateIncomingEpoch(payload.proof_proposal, currentEpoch);
+            await this.validatePreviousEpochRecordHash(payload.proof_proposal, currentEpoch);
+            const vdfParams = await this._state.requireSignedVDFParams();
+            await this.validateProofProposalVdfParametersHash(payload.proof_proposal, vdfParams);
+            await this.validateProofProposalVdfProof(payload.proof_proposal, vdfParams);
```
leonardostsouza, line 52: Just some "food for thought":

This is unlikely, but not impossible that the state might change while doing the validation, as we do several disconnected reads in this function. 

Can we implement an atomic read here somehow? Or at least mitigate this possibility

src/core/consensus/v1/validators/V1BaseConsensusOperation.js
b-garbacz: Ok I will normalize it

src/core/consensus/v1/validators/V1EpochProofProposalRequest.js
b-garbacz: Well, this references to the simulator idea also for validation layer for validators.

leonardostsouza (approved):
LGTM

PR #716, opened 2026-07-20
603 indexermanager implement proof propagator

leonardotc:


lucasfernandes:
This is duplicated and is being solved on #739

Issue #728, opened 2026-07-22
chore: improve consensus config storage


b-garbacz:
# Description 
Current VDF configuration storage is specific domain, and it might be a disadvantage in the future if we will decide to perform a transaction from autobase quorum consensus  + PoT -> PoS/Rotating indexer/Probabilistic Quorum, then VDF domain config will limit us and be redundant historical data. Also current implementation is not giving to us flexibility. 

# Task

Implement generic config storage

PR #736, opened 2026-07-22
chore: improve consensus config storage (#728)

b-garbacz:
## Description

Updates the current epoch consensus mechanism to support a generic format for storing consensus configurations.

## Motivation

To support different consensus rules and mechanism versions in the future, the ledger needs a flexible configuration format. We therefore introduced a generic schema instead of coupling the stored data to the current VDF requirements, such as difficulty and discriminant bit size:

```js
{
    sv: uint8 // Schema version
    cd: bytes // Encoded consensus configuration data
}
```

The configuration data can use any format (although size is limited to 3KB). The schema version determines which decoder is used to interpret it.

This PR adapts the current epoch consensus mechanism to use schema version 1.


cli/commandHandler.js
```diff
@@ -49,6 +55,7 @@ export class CommandHandler {
     #handlers;
     #pendingConfirmation = null;
     #pendingGenesisInitialization = null;
+    #pendingSetConsensusConfig = null;
```
leonardotc, line 58: This is getting pretty out of hand. If this needs state carried over between commands, it should have a shared context. If its a multiple step operation, it should have a coherent evaluate condition on top of this shared context. If we need a variable for each command, this means it wont scale.

src/core/state/State.js
```diff
@@ -612,34 +625,49 @@ class State extends ReadyResource {
         return this.#bee;
     }
 
-    async getSignedVDFParams() {
-        const vdfParamsBuffer = await this.getSigned(EntryType.VDF_PARAMS);
-        if (_.isNil(vdfParamsBuffer)) return null;
+    async getSignedConsensusConfig() {
+        const decodeConfigAtPointer = async currentConfigPointer => {
```
leonardotc: Why are using a function variable here?

leonardotc (changes requested):

src/core/state/State.js
b-garbacz: I used a local helper to separate the decoding logic. However, since it is used only once, we can inline it. What do you think about this?

leonardotc: it is probably better, its harder to read and to parse and that switch case will only grow bigger.

cli/commandHandler.js
leonardostsouza: I agree. I created a task to handle this and assigned P2 priority:
https://github.com/Trac-Systems/main_settlement_bus/issues/773

leonardostsouza: For now, i recommend keeping the code as it is and do the bigger refactor in a separate task

src/core/state/State.js
b-garbacz: Oki. Done.

leonardotc (approved):

Issue #733, opened 2026-07-22
chore: update StateValidationSchema to handle correctly new generic consensus config


b-garbacz:
Self described task

PR #738, opened 2026-07-23
chore: update StateValidationSchema to handle correctly new generic consensus config (#733)

b-garbacz:
## What

  - Added validation schema for the new generic SET_CONSENSUS_CONFIG operation.
  - Added nested validation for cco.cc, including sv and cd.
  - Limited cd to 1–3072 bytes.
  - Restricted sv to a non-zero single-byte value (1–255).
  - Updated fixtures and validation tests.

  ## Why

  To replace VDF-specific validation with a generic, versioned consensus configuration format while keeping operation payloads bounded and structurally safe.


proto/applyOperations/messages/set_consensus_config_operation.proto
```diff
@@ -3,14 +3,14 @@ syntax = "proto3";
...
+    consensusConfig cc    = 3; // consensus configuration
+    bytes in              = 4; // Nonce of the invoker (admin)
+    bytes is              = 5; // Signature of the invoker (admin)
 }
 
 message consensusConfig {
-  bytes version    = 1; // uint8
-  bytes configData = 2;
+  bytes sv    = 1; // schema version - uint8
+  bytes cd    = 2; // config data - generic bytes
```
leonardostsouza, line 15: Can you rebase and update the names in `applyOperationCodecs.js`? Otherwise, it will break the current implementation

leonardostsouza: Also, update the tests

leonardostsouza (approved):

Issue #595, opened 2026-05-27
Implement apply contract to handle time consensus epoch


b-garbacz:

leonardostsouza:
Some of the improvements will be delegated to another task: https://github.com/Trac-Systems/main_settlement_bus/issues/764

PR #739, opened 2026-07-24
feat: implement coordinator service

lucasfernandes:
This is the implementation for the state machine according to the diagram.
There was a huge impact on the network because it's required to reproduce the logic of connection to indexers.
There's an attempt to refactoring so both indexers and validators connections could live together and avoid the necessity of duplicated code. Hooking up the events for indexer promotion and epoch generation are on the scope of another task. We also provided a minimum implementation for set epoch apply, in order to be able to test it.

NOTE: There should be a VDF retry mechanism in case it returns an error. This should be verified during PR review


src/core/consensus/protocols/ConsensusMessages.js
```diff
@@ -10,12 +13,34 @@ class ConsensusMessages {
         this.#consensusRouter = new ConsensusRouterV1(state, wallet, config, pendingRequestService);
     }
 
-    setupProtomuxMessages(connection) {
-        connection.consensusProtocolSession = new ConsensusV1Protocol(
+    createProtomux(connection) {
```
leonardostsouza: This is a minor observation, but this function name is a bit misleading, as it doesn't create a Protomux instance. Maybe it would be best to rename it to `createProtocolSession` or something like this.

leonardostsouza: Specially because `const mux = Protomux.from(connection);` is in another function, not here

src/core/consensus/services/EpochCoordinatorOperations.js
```diff
@@ -0,0 +1,133 @@
...
+                approval_sig: signature,
+            })
+        );
+        const message = await applyStateMessageFactory(this.#wallet, this.#config)
+            .buildCompleteSetEpochMessage(this.#wallet.address, proofData, approvals);
+        return safeEncodeApplyOperation(message);
+    }
+
+    async appendSetEpoch(payload) {
+        return await this.#state.append(payload);
```
leonardostsouza: We don't need to `await` here

src/core/consensus/services/EpochCoordinatorOperations.js
```diff
@@ -0,0 +1,133 @@
...
+            );
+    }
+
+    async sendToIndexer(publicKeyHex, request, connectionManager) {
+        if (!connectionManager.connected(publicKeyHex)) throw new Error('Indexer not in the manager');
+
+        const response = await connectionManager.send(publicKeyHex, request);
+        return {
+            resultCode: response?.resultCode,
+            signature: response?.approval?.approval_sig,
```
leonardostsouza, line 63: If we are discarding `approval.approver` here and rebuilding it manually in `collectSignature`, then it is best to update the response schema definition (and maybe the class diagram) to reflect this. The indexers may return only a signature instead of sig + address. 

this can be done here or in a separate task. What do you think?

src/core/consensus/services/EpochCoordinatorOperations.js
```diff
@@ -0,0 +1,133 @@
...
+            throw err;
+        }
+
+        if (!memberSignature) throw new Error('Approval signature not received');
+
+        return {
+            signature: memberSignature,
+            approver: addressToBuffer(
+                publicKeyToAddress(publicKey, this.#config),
+                this.#config.addressPrefix
```
leonardostsouza: Can't we just reuse the constant `addressBuffer` here?

src/core/consensus/services/EpochCoordinatorOperations.js
```diff
@@ -0,0 +1,133 @@
...
+            network_id: proofProposal.network_id,
+            epoch: proofProposal.epoch,
+            previous_epoch_record_hash: proofProposal.previous_epoch_record_hash,
+            proposer: proofProposal.proposer,
+            vdf_parameters_hash: proofProposal.vdf_parameters_hash,
+            vdf_proof: proofProposal.vdf_proof,
+            signature: proofProposal.signature,
+        });
+        const approvals = signatures.map(({ signature, approver }) =>
+            safeEncodeProofProposalApproval({
```
leonardostsouza: We should not use safe encoder here. This function should throw in case of error

src/core/consensus/services/EpochCoordinatorOperations.js
```diff
@@ -0,0 +1,133 @@
...
+                approver,
+                approval_sig: signature,
+            })
+        );
+        const message = await applyStateMessageFactory(this.#wallet, this.#config)
+            .buildCompleteSetEpochMessage(this.#wallet.address, proofData, approvals);
+        return safeEncodeApplyOperation(message);
```
leonardostsouza: We should not use safe encoder here. This function should throw in case of error

src/core/consensus/services/EpochCoordinatorOperations.js
```diff
@@ -0,0 +1,133 @@
...
+            signature: memberSignature,
+            approver: addressToBuffer(
+                publicKeyToAddress(publicKey, this.#config),
+                this.#config.addressPrefix
+            )
+        };
+    }
+
+    async buildSetEpochPayload(proofProposal, signatures) {
+        const proofData = safeEncodeProofProposal({
```
leonardostsouza: We should not use safe encoder here. This function should throw in case of error

tests/unit/state/apply/setEpoch/setEpochScenarioHelpers.js
```diff
@@ -0,0 +1,186 @@
...
+    const adminNode = context.adminBootstrap;
+    const txValidity = await deriveIndexerSequenceState(adminNode.base);
+    const payload = await applyStateMessageFactory(adminNode.wallet, config)
+        .buildCompleteSetGenesisEpochMessage(
+            adminNode.wallet.address,
+            txValidity,
+            uint32ToBuffer(VDF_DIFFICULTY),
+            uint16ToBuffer(VDF_DISCRIMINANT_SIZE)
+        );
+    await appendAndUpdate(adminNode.base, safeEncodeApplyOperation(payload));
```
leonardostsouza: If the fixtures are wrong, we should throw here before even running the tests. Don't use a safe encoder here

src/core/state/State.js
```diff
@@ -3475,13 +3483,179 @@ class State extends ReadyResource {
         return Status.SUCCESS;
     }
 
-    async #handleApplySetEpochOperation(op, view, base, node, _batch) {
+    async #handleApplySetEpochOperation(op, view, base, node, batch) {
         if (!this.#stateValidationSchema.validateSetEpochOperation(op)) {
```
b-garbacz, line 3519: As we discussed last time - we should add here an information that this is placeholder

tests/unit/state/apply/setEpoch/setEpochScenarioHelpers.js
```diff
@@ -0,0 +1,186 @@
...
+                vdfSolution,
+                proofProposal.signature,
+                ConsensusResultCode.OK,
+                approverNode.wallet.address
+            );
+        approvals.push(responseMessage.proof_proposal_response.approval);
+    }
+
+    const proofData = safeEncodeProofProposal(proofProposal);
+    const encodedApprovals = approvals.map(approval => safeEncodeProofProposalApproval(approval));
```
leonardostsouza: Do not use safe encoders here (lines 175 and 176). We need to throw in case something goes wrong

tests/unit/state/apply/setEpoch/setEpochScenarioHelpers.js
```diff
@@ -0,0 +1,186 @@
...
+
+    const message = await applyStateMessageFactory(proposerNode.wallet, config)
+        .buildCompleteSetEpochMessage(proposerNode.wallet.address, proofData, encodedApprovals);
+
+    return safeEncodeApplyOperation(message);
+}
+
+export function decodeSetEpochPayload(payload) {
+    return safeDecodeApplyOperation(payload);
```
leonardostsouza: Do not use safe encoder. We need to throw in case something goes wrong

src/core/consensus/services/EpochCoordinatorService.js
```diff
@@ -0,0 +1,376 @@
...
+
+        // kick off the process
+        await machine.enter({ next: scheduleNext });
+    }
+
+    async #handleLoadEpochContext(_context, machine) {
+        // "snapshot"
+        const currentEpoch = await this.#state.getCurrentEpoch();
+        const currentEpochHash = await this.#state.getEpoch(currentEpoch);
+        const { vdfDifficulty, vdfDiscriminantSize } = await this.#state.getSignedVDFParams();
```
leonardostsouza: I'm thinking if we shouldn't use `requireCurrentEpoch()`, `requireEpoch()` and `requireSignedVDFParams()` here to throw in case the read returns `null`.

What do you think? Overkill?

src/core/consensus/services/EpochCoordinatorService.js
```diff
@@ -0,0 +1,376 @@
...
+            await machine.send(EPOCH_EVENTS.EXTERNAL_APPROVALS_REQUIRED);
+        }
+    }
+
+    async #handleSendProposalToIndexers(context, machine) {
+        const approvers = await this.#operations.approvers();
+        await this.#manager.connect();
+
+        if (approvers.length + 1 < machine.context.quorum) {
+            await machine.send(EPOCH_EVENTS.APPROVAL_COLLECTION_FAILED);
```
leonardostsouza, line 172: This event is not declared as a valid transition in `EpochStateMachine.js` for this state:

```js
 [EPOCH_STATES.SEND_PROPOSAL_TO_INDEXERS]: Object.freeze({
        [EPOCH_EVENTS.PROPOSAL_BROADCAST]: EPOCH_STATES.COLLECT_APPROVALS,
    }),
```

src/utils/StateMachine.js
```diff
@@ -0,0 +1,106 @@
...
+
+        this.#transitions = transitions;
+        this.#state = initialState;
+        this.#discardState = discardState;
+        this.#handlers = handlers;
+    }
+
+    async _close() {
+        this.clearListeners();
+    }
```
leonardostsouza, line 27: We should ensure that the state machine cannot change states after closed. Idk if only clearing the listeners is enough to do it. 

For example, consider this scenario:

```md
1. state machine enters `COLLECT_APPROVALS` and sends approval requests to several indexers.

2. requests remain pending while machine waits for quorum.

3. Another indexer creates the canon epoch. Local state emits `EPOCH_CREATED`.

4. coordinator closes the current machine (basically remove listeners only) and schedules a new cycle.

5. Even with the machine closed, the old pending approval requests are still valid (in theory) and the machine is still able to to transition state (also, in theory).

6. A pending approval from the abandoned cycle arrives after the machine has closed.

7. The approval callback still holds references to the closed machine and its proposal context.

8. The callback adds the approval and calls `machine.send(QUORUM_REACHED)`.

9. The closed machine cycle remains in `COLLECT_APPROVALS` . It transitions to `BUILD_SET_EPOCH`.

10. The "closed" machine resumes executing handlers and may build or append an epoch operation without its original listeners and timeouts.
```

Is this prevented somehow just by clearing listeners? Maybe we should at least transition the machine to a `CLOSED` state, from where it cannot transition. What do you think?

src/core/network/Network.js
```diff
@@ -97,72 +86,153 @@ class Network extends ReadyResource {
...
 
         this.cleanupPendingConnections();
-        await this.#validatorHealthCheckService.close();
-        await this.#epochProofProposalService.close();
         this.#validatorPendingRequestService.close();
         this.#transactionCommitService.close();
         this.#indexerPendingRequestService.close();
-
+        await this.#indexerConnectionManager.close();
+        await this.#validatorConnectionManager.close()
```
leonardostsouza, line 170: We should probably call `this.#epochCoordinatorService.close();` here too

src/core/network/services/ValidatorConnectionManager.js
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+
+    remove(publicKey) {
+        const publicKeyHex = this._toHexString(publicKey);
+        if (!this.exists(publicKeyHex)) return;
+
+        this.#stopHealthCheck(publicKeyHex);
+        this._connections.delete(publicKeyHex);
+    }
```
leonardostsouza, line 103: This one came from copilot:

```md
`ValidatorConnectionManager.remove()` is used for two different operations:

1. Detaching the validator role while keeping the shared socket open—for example, when the peer becomes an indexer.
2. Removing an unhealthy validator, where the connection should be closed or quarantined.

Currently, `remove()` always deletes only the validator entry and leaves the socket open. As a result, an unhealthy or rotated validator can be immediately added again using the same connection.

Additionally, `remove()` does not verify which connection is being removed. If an old duplicate connection closes, it can remove a newer valid connection registered for the same peer.
```

Might be overkill, but I wanted to put it here for your consideration.

src/core/consensus/services/EpochCoordinatorService.js
```diff
@@ -0,0 +1,376 @@
...
+        await super._close();
+        await this.#vdfService?.close();
+    }
+
+    async start() {
+        if (this.isSchedulerRunning) {
+            return false;
+        }
+
+        const started = super.start(this.#intervalMs);
```
b-garbacz: Can be inlined

src/core/network/protocols/NetworkMessages.js
```diff
@@ -42,7 +42,7 @@ class NetworkMessages {
         );
     }
 
-    setupProtomuxMessages(connection) {
+    createProtomux(connection) {
```
leonardostsouza: I made another commentary in a similar change.

src/core/consensus/services/EpochCoordinatorService.js
```diff
@@ -0,0 +1,376 @@
...
+            await machine.send(EPOCH_EVENTS.APPROVAL_COLLECTION_FAILED);
+            return;
+        }
+
+        const proposals = {
+            approvers, approvals: [], rejections: [], closed: false,
+        };
+        
+        machine.appendContext({ proposals });
+        this.#dispatchApprovalRequests(approvers, this.#manager, machine, proposals);
```
leonardostsouza, line 181: We try a connection on line 165, but it is still "pending", right? Maybe we should wait until the connection is successful (or timeout) before trying to dispach requests here.

src/core/consensus/services/EpochCoordinatorOperations.js
leonardotc: ok

src/core/network/protocols/NetworkMessages.js
leonardostsouza: https://github.com/Trac-Systems/main_settlement_bus/pull/739#discussion_r3734288652

src/core/consensus/services/EpochCoordinatorService.js
```diff
@@ -0,0 +1,376 @@
...
+        if (this.isSchedulerRunning) {
+            return false;
+        }
+
+        const started = super.start(this.#intervalMs);
+        return started;
+    }
+
+    async stop(waitForCurrent = true) {
+        return super.stop(waitForCurrent);
```
b-garbacz, line 65: `stop()` does not cancel the active epoch round or close the VDF service.

src/core/network/Network.js
b-garbacz: +1

`#epochCoordinatorService` should also close the VDF service and active epoch round

src/core/consensus/services/EpochCoordinatorService.js
```diff
@@ -0,0 +1,376 @@
...
+        machine.appendContext({ proposals });
+        this.#dispatchApprovalRequests(approvers, this.#manager, machine, proposals);
+        await machine.send(EPOCH_EVENTS.PROPOSAL_BROADCAST);
+    }
+
+    #dispatchApprovalRequests(approvers, indexerConnectionManager, machine, proposals) {
+        for (const member of approvers) {
+            const {currentEpoch, currentEpochHash, vdf} = machine.context;
+            this.#operations.collectSignature(member, {currentEpoch, currentEpochHash, vdf}, indexerConnectionManager)
+                .then((confirmation) => this.#handleApproval(confirmation, proposals, machine))
```
b-garbacz, line 189: We send an approval request and register this callback. Before the peer responds, and let's assume the epoch round is closed or replaced. The peer responds later and this callback is executed. The callback uses the old machine and calls `machine.send(QUORUM_REACHED)`, however send() does not check whether the machine is closed, so the old round can continue executing its handlers.

As a result, a finished round can be resumed by a late approval. Could we cancel pending requests when closing a round and prevent `send()` from processing events after `close()`? 

Maybe we could cancel all pending approval requests when the round is closed. This would prevent late responses from continuing the old round.

b-garbacz: Btw you could write a regression test for both `EPOCH_CREATED` and `stop()`

src/core/consensus/services/EpochStateMachine.js
```diff
@@ -0,0 +1,103 @@
...
+        [EPOCH_EVENTS.APPEND_FAILED]: EPOCH_STATES.BACKOFF,
+        [EPOCH_EVENTS.TARGET_EPOCH_ALREADY_SIGNED]: EPOCH_STATES.RELOAD_SIGNED_CONTEXT,
+    }),
+    [EPOCH_STATES.REFRESH_SIGNED_STATE]: Object.freeze({
+        [EPOCH_EVENTS.EPOCH_UNCHANGED_WITHOUT_LOCAL_PROOF]: EPOCH_STATES.INITIALIZE_VDF,
+        [EPOCH_EVENTS.EPOCH_UNCHANGED_WITH_LOCAL_PROOF]: EPOCH_STATES.BUILD_AND_SIGN_PROOF_PROPOSAL,
+        [EPOCH_EVENTS.NEW_EPOCH_DISCOVERED]: EPOCH_STATES.RELOAD_SIGNED_CONTEXT,
+    }),
+    [EPOCH_STATES.BACKOFF]: Object.freeze({
+        [EPOCH_EVENTS.BACKOFF_ELAPSED]: EPOCH_STATES.REFRESH_SIGNED_STATE,
```
b-garbacz, line 87: Image

We have the `BACKOFF -> REFRESH_SIGNED_STATE` transition in both the diagram and EpochStateMachine, but it is never used.

`Autobase` has a method `await base.update()`, which fetch data from other peers and update linearizer. But this means more like "process all available data from other nodes".

src/core/state/State.js
b-garbacz, line 1: The coordinator reads the VDF parameters when the round starts and uses them to calculate the proof. If `SET_VDF_PARAMS` changes these parameters during the round, the coordinator is not notified. This logic should be handled here regardless of whether we use `SET_VDF_PARAMS` or the generic consensus configuration from the chore branch that we worked on with @leonardostsouza. In either case, we need to handle this scenario here.

src/core/consensus/services/EpochCoordinatorOperations.js
lucasfernandes: Yes, updated.

src/core/consensus/services/EpochCoordinatorOperations.js
lucasfernandes: Agreed, updated.

src/core/network/Network.js
```diff
@@ -97,72 +86,153 @@ class Network extends ReadyResource {
...
+
+        this.#validatorConnectionManager = new ValidatorConnectionManager(this.#config.maxValidators, this.#config, this.#logger, this.#networkMessages);
+        await this.#validatorConnectionManager.ready();
+
+        this.#validatorMessageOrchestrator = new MessageOrchestrator(this.#validatorConnectionManager, this.#state, this.#config, this.#wallet);
+
+        this.#consensusMessages = new ConsensusMessages(this.#state, this.#wallet, this.#config, this.#indexerPendingRequestService);
+
+        const indexerCount = await this.#state.indexerCount();
+        this.#indexerConnectionManager = new IndexerConnectionManager(
```
b-garbacz, line 132: What if quorum will change later? Maybe we should have a call that updates this limit.

src/core/network/Network.js
```diff
@@ -97,72 +86,153 @@ class Network extends ReadyResource {
...
+        const stream = this.#store.replicate(connection);
+        wakeup.addStream(stream);
+    }
 
-        this.#state.on(CustomEventType.IS_NON_INDEXER, async (publicKey) => {
-            const indexersCount = await this.#state.indexerCount()
-            this.#indexerConnectionManager.setMax(indexersCount);
-            this.disconnectIndexerPeer(publicKey, 'peer demoted from indexer');
+    #listerners() {
+        this.#state.on(CustomEventType.IS_INDEXER, (publicKey) => {
```
b-garbacz, line 181: This event should enable coordinator, or we decided to do it in another task?

src/core/network/Network.js
```diff
@@ -97,72 +86,153 @@ class Network extends ReadyResource {
...
             this.disconnectValidatorPeer(publicKey, 'peer became unwritable');
         });
 
-        this.#state.on(CustomEventType.IS_INDEXER, bufferAddress => {
+        this.#state.on(CustomEventType.IS_NON_INDEXER, (bufferAddress) => {
             const address = tracCryptoApi.address.encode(this.#config.addressPrefix, bufferAddress);
             if (address === this.#wallet.address) {
-                this.#indexerObserverService.start();
-                this.#epochProofProposalService.start();
+                this.#epochCoordinatorService.stop();
```
b-garbacz, line 193: Question: The local node can still sign proposal approvals after losing it's indexer role?

src/core/consensus/services/EpochStateMachine.js
```diff
@@ -0,0 +1,103 @@
...
+    LOAD_EPOCH_CONTEXT: "LOAD_EPOCH_CONTEXT",
+    INITIALIZE_VDF: "INITIALIZE_VDF",
+    BUILD_AND_SIGN_PROOF_PROPOSAL: "BUILD_AND_SIGN_PROOF_PROPOSAL",
+    SEND_PROPOSAL_TO_INDEXERS: "SEND_PROPOSAL_TO_INDEXERS",
+    COLLECT_APPROVALS: "COLLECT_APPROVALS",
+    QUORUM_DECISION: "QUORUM_DECISION",
+    BUILD_SET_EPOCH: "BUILD_SET_EPOCH",
+    REFRESH_SIGNED_STATE_BEFORE_APPEND: "REFRESH_SIGNED_STATE_BEFORE_APPEND",
+    APPEND_SET_EPOCH: "APPEND_SET_EPOCH",
+    SEND_APPEND_SIGNAL: "SEND_APPEND_SIGNAL", // Terminal on purpose: no state is preserved across cycles, so there is nothing to hand off to BACKOFF.
```
b-garbacz, line 13: This is not used

src/core/consensus/services/EpochStateMachine.js
```diff
@@ -0,0 +1,103 @@
...
+    QUORUM_REACHED: "QUORUM_REACHED",
+    PEER_EPOCH_MISMATCH: "PEER_EPOCH_MISMATCH",
+    APPROVAL_COLLECTION_FAILED: "APPROVAL_COLLECTION_FAILED",
+    SET_EPOCH_BUILT: "SET_EPOCH_BUILT",
+    SIGNED_STATE_REFRESHED: "SIGNED_STATE_REFRESHED",
+    TARGET_EPOCH_ABSENT: "TARGET_EPOCH_ABSENT",
+    TARGET_EPOCH_ALREADY_SIGNED: "TARGET_EPOCH_ALREADY_SIGNED",
+    APPEND_ACCEPTED: "APPEND_ACCEPTED",
+    APPEND_FAILED: "APPEND_FAILED",
+    APPEND_SIGNAL_SENT: "APPEND_SIGNAL_SENT",
```
b-garbacz: Also this is not used

src/core/consensus/services/EpochStateMachine.js
```diff
@@ -0,0 +1,103 @@
...
+    PROPOSAL_BUILT: "PROPOSAL_BUILT",
+    EXTERNAL_APPROVALS_REQUIRED: "EXTERNAL_APPROVALS_REQUIRED",
+    EXTERNAL_APPROVALS_NOT_REQUIRED: "EXTERNAL_APPROVALS_NOT_REQUIRED",
+    PROPOSAL_BROADCAST: "PROPOSAL_BROADCAST",
+    APPROVAL_COLLECTED: "APPROVAL_COLLECTED",
+    QUORUM_REACHED: "QUORUM_REACHED",
+    PEER_EPOCH_MISMATCH: "PEER_EPOCH_MISMATCH",
+    APPROVAL_COLLECTION_FAILED: "APPROVAL_COLLECTION_FAILED",
+    SET_EPOCH_BUILT: "SET_EPOCH_BUILT",
+    SIGNED_STATE_REFRESHED: "SIGNED_STATE_REFRESHED",
```
b-garbacz: Also this and more...

src/core/consensus/services/EpochCoordinatorOperations.js
leonardotc: done

src/core/consensus/protocols/ConsensusMessages.js
leonardotc: done

src/core/consensus/services/EpochCoordinatorOperations.js
leonardotc: done

src/core/consensus/services/EpochCoordinatorOperations.js
leonardotc: done

src/core/consensus/services/EpochCoordinatorService.js
```diff
@@ -0,0 +1,376 @@
...
+        if (latestEpoch > context.currentEpoch) {
+            await machine.send(EPOCH_EVENTS.NEW_EPOCH_DISCOVERED);
+        } else if (context.vdf) {
+            await machine.send(EPOCH_EVENTS.EPOCH_UNCHANGED_WITH_LOCAL_PROOF);
+        } else {
+            await machine.send(EPOCH_EVENTS.EPOCH_UNCHANGED_WITHOUT_LOCAL_PROOF);
+        }
+    }
+
+    async #handleBackoff(context, _machine) {
```
b-garbacz: This flow looks different from the diagram. Didn't we want to refresh the signed state and either reuse the existing proof or reload the context, depending on whether it has changed???
Currently `context.next()` closes the current machine. The next scheduler cycle creates a new machine, so the proof stored in the current context is lost and VDF is calculated again.

b-garbacz (changes requested):

tests/unit/state/apply/setEpoch/setEpochScenarioHelpers.js
leonardotc: done

tests/unit/state/apply/setEpoch/setEpochScenarioHelpers.js
leonardotc: ok

src/core/consensus/services/EpochCoordinatorService.js
leonardostsouza: This sounds linked to my comment here:

https://github.com/Trac-Systems/main_settlement_bus/pull/739#discussion_r3735446498

src/core/consensus/services/EpochCoordinatorOperations.js
leonardostsouza: What I mean is, the ProposalApproval may consist only of a signature, instead of also including an `approver` field. What do you think?

image

src/core/consensus/services/EpochCoordinatorService.js
leonardotc: added the missing state. I think backoff logic is slightly incorrect since its equivalent to a "restart this", it shouldnt perform `next(aTimeThere)` but rather a `next(0)`. I am just going to check and push that.

src/core/network/Network.js
leonardotc: another task.

src/core/state/State.js
leonardotc: Iirc this was a working in progress. Is there, in this branch, a signal that allow us to be notified that the parameter changed?

src/core/consensus/services/EpochCoordinatorOperations.js
leonardotc: I guess it would work but it would be low-key weird since we would need to try verify signatures in the apply for every indexer until it returns true (which is dirty and would need extra guards to ensure uniqueness etc). We would also require to change that payload thingy which is in a different place. If the `trac-crypto-api` worked like `ecrecover` that returns the signer address, we could verify if that address is an indexer or something but afaik it returns `true` or `false`.

src/utils/StateMachine.js
leonardotc: This is done (the change aint in this line though).

tests/unit/state/apply/setEpoch/setEpochScenarioHelpers.js
leonardotc: done

src/core/network/Network.js
lucasfernandes: Agreed, fixed

src/core/consensus/services/EpochStateMachine.js
lucasfernandes: Fixed by the missing-state commit SEND_APPEND_SIGNAL is used now (transition in EpochStateMachine.js and handler in EpochCoordinatorService.js), with a passing test covering it.

src/core/consensus/services/EpochCoordinatorService.js
lucasfernandes: Agreed, fixed.

src/core/consensus/services/EpochStateMachine.js
lucasfernandes: removed

src/core/consensus/services/EpochStateMachine.js
lucasfernandes: removed

src/core/consensus/services/EpochCoordinatorService.js
lucasfernandes: Confirmed, fixed on @leonardostsouza comment: send() now no-ops if the machine is closing/closed, covered by a regression test.

src/core/network/services/ValidatorConnectionManager.js
lucasfernandes: Fixed part 2 (identity check), mirrors IndexerConnectionManager.remove(). Left part 1 out on purpose: one call site needs the socket kept open, so that's a separate change.

src/core/consensus/services/EpochCoordinatorService.js
lucasfernandes: Not overkill, I did it. requireCurrentEpoch() is a bit redundant there (already guarded earlier), but requireEpoch()/requireSignedVDFParams() had zero protection before, so those two are a real improvement.

src/core/consensus/services/EpochCoordinatorService.js
lucasfernandes: Confirmed and fixed, but not exactly at that line though, ended up in EpochCoordinatorOperations.sendToIndexer instead, since that's where the resolved publicKeyHex was already available. Now waits (up to connectTimeoutMs, polling every 50ms) for the connection before giving up.

src/core/consensus/services/EpochCoordinatorService.js
leonardotc: done

src/core/consensus/services/EpochCoordinatorService.js
leonardotc: We will probably need to talk about which extent the current epoch should be canceled. I think it will pretty hard to stop the current "step" on the process but we can prevent the current state machine to change state. We will implement the thing on the vdf.

src/core/consensus/services/EpochStateMachine.js
leonardotc: done

src/core/consensus/services/EpochCoordinatorOperations.js
leonardostsouza: Ok, then we are discarding `approver` field here, but it will still be used in apply function, right? 
Nevermind this commentary then

src/core/network/Network.js
lucasfernandes: This should be most likely barred in 2 places: 
1 - OperationHandler where it should verify if self is an indexer.
2 - In the Network where the service should remove the indexer from the connection manager.

I'm not sure who should implement the first one, but the second one is related to events handler in another task.

src/core/consensus/services/EpochCoordinatorService.js
leonardotc: test

src/core/consensus/services/EpochStateMachine.js
lucasfernandes: It's done -> This required a lot of rework. The reason for that is that there's a mismatch between the service lifecycle and the state machine inner working. For this particular case, the state machine will trigger a new state that will work as a blank slate for the service in question. So both lifecycles overlap.

src/core/network/Network.js
lucasfernandes: This should be in the event handler which is in a different task.

src/core/consensus/services/EpochCoordinatorService.js
lucasfernandes: Yes, we fixed. There's circumstances where the BACKOFF implies a new state machine. This is something to processed.

src/core/network/Network.js
b-garbacz: @lucasfernandes wdym? Do we have a ticket for this?

src/core/consensus/services/EpochStateMachine.js
b-garbacz: Actually this step was redundant, so +1

src/core/consensus/services/EpochStateMachine.js
b-garbacz: So the reason of the deletion is because it's duplicated right? We have already this https://github.com/Trac-Systems/main_settlement_bus/pull/739#discussion_r3750313360. Right?

src/core/state/State.js
leonardotc: Well, that is not exactly a placeholder. I think its just incomplete given the amount of rebases. The major issue is that comments with TODOs are extremely bad practice. Nobody tracks comments, we track tasks. Furthermore, someone can change the code and the comment could still be there but invalid. There is already a task to track the apply function completion.

src/core/state/State.js
```diff
@@ -3499,13 +3515,201 @@ class State extends ReadyResource {
...
+        }
+
+        const encodedEpochProof = safeEncodeEpochProof({ pd: op.seo.pd, app: op.seo.app });
+        if (encodedEpochProof.length === 0) {
+            this.#safeLogApply(OperationType.SET_EPOCH, "Failed to encode epoch proof.", node.from.key)
+            return Status.FAILURE;
+        }
+
+        const epochProofHash = await tracCryptoApi.hash.blake3Safe(encodedEpochProof);
+        const nextEpochBuffer = b4a.alloc(8);
```
leonardostsouza, line 3701: It would be good to check if epochProofHash actually contains something, as iblake3Safe returns an empty buffer in case of error

leonardostsouza: We should at least check if we got exactly 32 bytes

src/core/state/State.js
b-garbacz: It seems we don't have this signal, we should create it here or in a new task (maybe for improvements?)

leonardotc: New task, adapting the current state with the consensus stuff is already part of the integration. I would just change the description of tha task to include emitting an event.

src/core/network/Network.js
leonardotc: Not only we have the ticket, we already have the PR (which solves that). Its still not complete (and its rebased with a previous version of this): https://github.com/Trac-Systems/main_settlement_bus/pull/769 .

src/core/state/State.js
b-garbacz: Oki then resolved

src/core/state/State.js
lucasfernandes: I agree. Maybe we should let this to the #595. What do you think?

leonardostsouza: ok. No problem

leonardostsouza (approved):

b-garbacz (approved):

PR #741, opened 2026-07-28
chore: prepare MSB for pear runtime v3

b-garbacz:
- Added a runner that uses pear run on v2 and pear-runtime on v3.
- Routed all MSB startup scripts through the new runner.
- Added Bare.argv support for Pear v3 worker arguments.
- Replaced direct TTY streams with process stdio because v3 workers use piped streams.
- Added and pinned required runtime dependencies.
- Included required runtime files in the npm package.
- Added runner and argument compatibility tests.
- Updated the runtime documentation.

### DIFF
--- a/scripts/pear-runner.js
+++ b/scripts/pear-runner.js
@@ -37,13 +37,12 @@ export const detectPearMajor = (runCommand = spawnSync) => {
     }
 
     const output = `${result.stdout ?? ''}\n${result.stderr ?? ''}`;
-    const major = parsePearMajor(output);
 
-    if (result.status !== 0 || major === null) {
+    if (result.status !== 0) {
         throw new Error('Could not determine the installed Pear major version from `pear -v`.');
     }
 
-    return major;
+    return parsePearMajor(output);
 };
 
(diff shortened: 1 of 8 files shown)


scripts/pear-runner.js
```diff
@@ -0,0 +1,171 @@
...
+        runPearV3 = runWithPearV3
+    } = {}
+) => {
+    const major = detectVersion();
+
+    if (major !== null && major < 3) {
+        return runPearV2(args);
+    }
+
+    return runPearV3(args);
```
leonardostsouza, line 157: If major is `null` we are going to try running it with pear `v3`. We should return a message to tell the user to install `pear` instead

leonardostsouza: Currently we depend on `runPearV3` throwing an error, which is not ideal

b-garbacz: This is intentional because Pear v3 runs through the embedded `pear-runtime` dependency and does not require the Pear CLI anymore!

leonardostsouza (approved):

PR #742, opened 2026-07-28
chore: update cli for setting a new generic consensus config

leonardostsouza:


cli/commandHandler.js
```diff
@@ -423,7 +444,7 @@ export class CommandHandler {
         return "Set VDF discriminant bit size (example 2048):";
     }
 
-    #getSetVdfParamsDifficultyPrompt() {
-        return "Set new VDF difficulty (example 55_000_000):";
+    #getSetConsensusConfigPrompt() {
```
b-garbacz, line 447: I have one question:

Wouldn’t it be easier if the dialog asked the user only for `difficulty` and `dbs`, and then generated the config structure automatically based on those values?

The dialog could then display the generated structure and ask the user to confirm whether it looks correct. If the user confirms, we apply it; if not, we cancel.

In my opinion, it is easier and less error to provide two explicit values, let the CLI build the structure in the background, and then give the user a final confirmation step before saving it.

What do you think about this.

b-garbacz: Ofc if you prefer this way then it's fine. But from the user perspective brackets are confusing a lil bit

leonardostsouza: It would be easier, but then this would not be a generic configuration input method.

b-garbacz: Resolved on DM

b-garbacz (approved):

PR #747, opened 2026-07-30
Bound Corestore cache and shutdown replication

TracSystems:
This PR carries the Mayhem-tested trac-msb fixes for two production issues:

- bound the Corestore/Hyperbee cache size used by the MSB stack to avoid unbounded resident memory growth under sustained peer traffic;
- guard the Hyperswarm connection handler during shutdown so a late Pear v3 connection cannot call `store.replicate()` after Corestore has already started closing, which previously aborted Intercom with `Corestore is closed`.

Focused and full unit verification passed locally after the shutdown guard:

- `./node_modules/.bin/brittle-node -t 60000 tests/unit/network/Network.test.js`
- `npm run test:pear-runner`
- `npm run test:unit:node` (`1011/1011` tests, `26545/26545` asserts)


src/config/config.js
```diff
@@ -2,6 +2,8 @@ import b4a from 'b4a'
 import { isDefined } from '../utils/type.js'
 import _ from 'lodash'
 
+const DEFAULT_HYPERBEE_CACHE_MAX_ENTRIES = 65_536
```
b-garbacz, line 5: We generally keep environment config as the single source of truth, so I’d remove this additional default and use config.hyperbeeCacheMaxEntries directly.

src/config/config.js
```diff
@@ -101,6 +104,18 @@ export class Config {
         return this.#config.maxRetries
     }
 
+    get hyperbeeCacheMaxEntries() {
+        const value = this.#isOverriden('hyperbeeCacheMaxEntries')
+            ? this.#options.hyperbeeCacheMaxEntries
+            : this.#config.hyperbeeCacheMaxEntries ?? DEFAULT_HYPERBEE_CACHE_MAX_ENTRIES
+
+        if (!Number.isSafeInteger(value) || value < 1) {
```
b-garbacz, line 112: Let’s use _.isSafeInteger(value) here for consistency with the rest of the class.

TracSystems:
Closing this branch because later commits broadened into storage normalization/wrapper behavior. Will resubmit only narrowly proven fixes from the release baseline.

PR #753, opened 2026-07-30
chore: create interface for direct set vdf params in cli

leonardostsouza:


cli/commandHandler.js
```diff
@@ -40,7 +40,8 @@ export const COMMANDS = {
     GET_TX_DETAILS: "/get_tx_details",
     GET_EXTENDED_TX_DETAILS: "/get_extended_tx_details",
     EPOCH_GENESIS_INITIALIZATION: "/init_genesis",
-    SET_CONSENSUS_CONFIG: "/set_consensus_config"
+    SET_CONSENSUS_CONFIG: "/set_consensus_config",
+    SET_VDF_PARAMS: "/set_vdf_params"
```
b-garbacz, line 44: Are we going to keep them both?

b-garbacz (approved):
lgtm

cli/commandHandler.js
leonardostsouza: Yes. The plan was to have a generic `/set_consensus_config` and a simpler interface for `/set_vdf_params`. Internally, they both call the same handler, though

PR #768, opened 2026-08-06
fix: guard network replication during shutdown

TracSystems:
## Summary\n- prevent late Hyperswarm connection handlers from calling Corestore.replicate after MSB shutdown begins\n- reject pending requests and destroy late/failed connections instead of letting a closed Corestore exception abort the runner\n- add a regression test for the late-connection/shutdown race\n\n## Scope\nThis is the MSB-side shutdown guard only. It does not change Autobase, Hypercore, Corestore storage semantics, consensus data, signatures, DHT behavior, public MSB behavior, or protocol formats.\n\n## Verification\n- npm run test:unit:node: 1010/1010 pass\n- npm run test:pear-runner: 8/8 pass

### DIFF
--- a/src/core/network/Network.js
+++ b/src/core/network/Network.js
@@ -177,8 +177,12 @@ class Network extends ReadyResource {
         store,
         wallet,
     ) {
+        this.#assertCanReplicate();
+
         if (!this.#swarm) {
             const { wallet: wrappedWallet, keyPair } = await this.#getOrGenerateWallet(store, wallet);
+            this.#assertCanReplicate();
+
             this.#wallet = wrappedWallet
             this.#validatorMessageOrchestrator.setWallet(this.#wallet);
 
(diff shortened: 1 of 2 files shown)


b-garbacz (approved):

leonardotc (changes requested):
There is a flag for that already does  that on ready resource.

```
  constructor () {
    super()
    this._id = nextId++;

    this.opening = null
    this.closing = null

    this.opened = false
    this.closed = false
  }
```

One of those should do the trick.

src/core/network/Network.js
```diff
@@ -104,8 +107,15 @@ class Network extends ReadyResource {
         this.#pendingRequestsService.close();
         this.#transactionCommitService.close();
 
-        if (this.#swarm !== null) {
-            this.#swarm.destroy();
+        const swarm = this.#swarm;
+        if (swarm !== null) {
+            if (typeof swarm.removeAllListeners === 'function') {
```
leonardotc, line 112: This looks vibecoded. remove it.

leonardotc (changes requested):

leonardotc (approved):

src/core/network/Network.js
b-garbacz: chore: clean up network module with redundant attributes and better error handling in connection section

PR #769, opened 2026-08-06
feat: hook up network events

leonardotc:
- This hooks the external events for the indexer manager and service associated with the epoch. It changes according the hooks on the network js.
- A refactor of the cli started here since it grew quite during the feature and it wasnt scale. The current state is (obviously) not final. It scales better but there is an implicit order to be followed, there is no structural correlation between steps of the same operation (e.g.: they are not in an array) and there are still implicit steps build as sub-operation (those weird callbacks) that shouldnt be there.

### DIFF
--- a/cli/commandHandler.js
+++ b/cli/commandHandler.js
@@ -8,6 +8,17 @@ import {
     MAX_VDF_DISCRIMINANT_BIT_SIZE
 } from "../src/utils/constants.js";
 
+const OPERATIONS = {
+    CONFIRMATION: "confirmation",
+    GENESIS_DIFFICULTY: "genesis-difficulty",
+    GENESIS_DISCRIMINANT_BIT_SIZE: "genesis-discriminant-bit-size",
+    CONSENSUS_CONFIG: "consensus-config",
+    VDF_DIFFICULTY: "vdf-difficulty",
+    VDF_DISCRIMINANT_BIT_SIZE: "vdf-discriminant-bit-size",
+    GENESIS_PREFIX: "genesis-",
+    VDF_PREFIX: "vdf-"
+};
+
 export const COMMANDS = {
     HELP: "/help",
     EXIT: "/exit",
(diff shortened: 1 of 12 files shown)


cli/commandHandler.js
```diff
@@ -65,35 +61,35 @@ export class CommandHandler {
...
+            {
+                evaluate: ({ context }) => context.pending?.op === "confirmation",
+                process: async ({ input, context }) => this.#handlePendingConfirmation(input, context)
+            },
+            {
+                evaluate: ({ context }) => context.pending?.op.startsWith("genesis-"),
+                process: async ({ input, context }) => this.#handlePendingGenesisInitialization(input, context)
+            },
+            {
+                evaluate: ({ context }) => context.pending?.op === "consensus-config",
```
jusufsuljic: Small thing, but can we use constants in this case? I mean so we contain these strings somewhere, especially if they will be used more than just here/once.

src/core/network/Network.js
```diff
@@ -178,19 +180,40 @@ class Network extends ReadyResource {
     }
 
     #listerners() {
```
jusufsuljic: We should fix this typo, it looks old.

leonardotc: Yea, look like an antiseptic brand. I will change it.

cli/commandHandler.js
leonardotc: Yeah, those strings are temporary but I should change to a single const, yes.

lucasfernandes (approved):
LGTM

jusufsuljic (approved):

src/core/network/Network.js
b-garbacz, line 89: dead code

src/core/network/Network.js
```diff
@@ -177,20 +179,41 @@ class Network extends ReadyResource {
...
+                if (await this.#state.getCurrentEpoch() !== null) {
+                    this.#epochCoordinatorService.start();
+                }
+            }
+        });
+
+        this.#state.on(CustomEventType.IS_NON_INDEXER, async (bufferAddress) => {
             const address = tracCryptoApi.address.encode(this.#config.addressPrefix, bufferAddress);
             if (address === this.#wallet.address) {
                 this.#epochCoordinatorService.stop();
```
b-garbacz: 1. missing await
2. State machine is still active

lucasfernandes: added await

src/core/network/Network.js
leonardotc: removed

src/core/network/Network.js
b-garbacz: Oki,how about state machine?

leonardotc: second item is valid imo. i made the state machine detached from the object (it aint an instance etc) to enforce context isolation. I am probably gonna switch to event-driven approach to stop the thing.

b-garbacz (approved):

PR #770, opened 2026-08-06
feat: fix v1 result codes

### LINKED ISSUE #694: Fix returned result codes in V1
### Description
Check returned result codes in V1 Network Protocol and update returned result codes when errors happen.

For example, there are instances of failed schema validation in which the code throws a `INVALID_PAYLOAD` instead of `SCHEMA_VALIDATION_FAILED`

Note: The scope for this task is V1 Network Protocol only. Both Network and Epoch Consensus protocols share several error codes, but epoch consensus protocol is out of scope for this task

### Objectives
- [x] Look for mismatched result codes
- [x] Replace wrong result codes with the most relevant for each situation
- [x] Update unit tests

jusufsuljic:


leonardostsouza (approved):
LGTM

jusufsuljic:
On hold until we merge the epoch consensus PR.

PR #776, opened 2026-08-11
chore: Replace vdfParamsHash with VDF params

b-garbacz:
## Summary

This PR replaces the 32 byte vdf_parameters_hash with explicit VDF parameters:

- difficulty 4 bytes
- discriminant_bit_size 2 bytes

This saves 26 bytes of raw parameter data per proposal and removes additional hashing operation hashing.

Additionally, it:
- Supports variable VDF proof sizes, so the proof size is no longer fixed at 516 bytes and now depends on the selected discriminant size.
- Restricts discriminant sizes to 1024, 2048, and 4096 bits. Unsupported values are rejected across the API, CLI, consensus validation, and state validation layers.
- Validates the corresponding proof length. Each supported discriminant requires an exact proof size: 260, 516, or 1028 bytes respectively.
- Difficulty and discriminant_bit_size must match the configuration stored in the signed state, preventing proposals from using different or weaker VDF
  parameters.
- Payloads containing missing, duplicated, empty, zero-filled, incorrectly encoded, or unknown protobuf fields are rejected before they can affect the state.
- Fixed problem related to address validation.


jusufsuljic (approved):

lucasfernandes (approved):
LGTM

PR #780, opened 2026-08-18
feat: update apply contract to handle time consensus epoch

leonardostsouza:


src/core/state/State.js
```diff
@@ -3527,6 +3529,11 @@ class State extends ReadyResource {
...
         }
 
+        if (!b4a.equals(op.address, proofProposal.proposer)) {
+            this.#safeLogApply(OperationType.SET_EPOCH, "Operation address does not match the proof proposer.", node.from.key)
+            return Status.FAILURE;
+        }
+
         const currentEpochBuffer = await this.#getEntryApply(EntryType.EPOCH_CURRENT, batch);
         if (currentEpochBuffer === null) {
             this.#safeLogApply(OperationType.SET_EPOCH, "Current epoch is not initialized. Genesis epoch has not been set.", node.from.key)
```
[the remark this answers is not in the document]
leonardostsouza: longer/shorter value would imply corrupted state. This length is checked before appending.

src/core/state/State.js
```diff
@@ -3564,17 +3570,13 @@ class State extends ReadyResource {
...
             this.#safeLogApply(OperationType.SET_EPOCH, "Consensus config is not initialized.", node.from.key)
             return Status.FAILURE;
         }
 
-        const currentConsensusConfigIndex = safeReadUint32BE(currentConsensusConfigBuffer);
-        if (currentConsensusConfigIndex === null) {
-            this.#safeLogApply(OperationType.SET_EPOCH, "Failed to read current consensus config index from buffer", node.from.key)
-            return Status.FAILURE;
-        }
+        const currentConsensusConfigIndex = currentConsensusConfigIndexBuffer.readUInt32BE(0);
```
jusufsuljic: Why not keep safeReadUint32BE?

leonardostsouza: My reasoning:
- This value is checked before appending it to the ledger.
- Under no valid circumstance will this value be null, but malformed
- Thus, an exception being thrown by readUInt32BE means the local state is corrupted. So I preferred to just halt execution in this case.

leonardostsouza: My reasoning:
- This value comes directly from the ledger, not from payload
- This value is checked before being appended it to the ledger, so it can be trusted
- We check for `null` beforehand, and no valid circumstance will this value be malformed (namely, wrong length)
- Thus, an exception being thrown by readUInt32BE means the local state is corrupted. So I preferred to just halt execution in this case.

leonardostsouza: Coming to think of it, this will return an ugly exception instead of a nice error message. I am going to adjust it

src/core/consensus/services/EpochCoordinatorService.js
```diff
@@ -344,7 +344,26 @@ class EpochCoordinatorService extends SchedulableService {
 
                 const targetEpoch = stateMachine.context.currentEpoch + 1n;
                 stopAppendListener = listenTo(this.#state, CustomEventType.EPOCH_CREATED, ({ epoch, proposerAddress }) => {
-                    if (epoch !== targetEpoch) return;
+                    if (
```
leonardotc: Why is this required? Internal structure details like this shouldnt leak to high order services. Both this epoch and the target one should be a resolved number string etc. This is an internal component, you can rely (and should rely) on its interface.

src/core/state/State.js
```diff
@@ -3532,18 +3538,28 @@ class State extends ReadyResource {
             this.#safeLogApply(OperationType.SET_EPOCH, "Current epoch is not initialized. Genesis epoch has not been set.", node.from.key)
             return Status.FAILURE;
         }
+        
+        const nextEpochBuffer = incrementBuffer(currentEpochBuffer, EPOCH_BYTE_LENGTH);
```
leonardotc, line 3557: I think there is some to add buffers in the balance. Cant reuse here?

src/utils/buffer.js
```diff
@@ -200,3 +200,78 @@ export function isZeroBuffer(buffer) {
...
+
+/**
+ * Converts a big-endian unsigned integer buffer to a numeric string.
+ * The input must contain between 1 and 16 bytes.
+ *
+ * @param {Buffer} buffer - The unsigned integer to convert.
+ * @param {'decimal'|'hex'} [encoding='decimal'] - Output encoding.
+ * @returns {string|null} The normalized numeric string, or null for invalid input.
+ */
+export function toUIntString(buffer, encoding = 'decimal') {
```
leonardotc, line 248: I think the standard is to create a version of this that throws an exception and then make it "safe".

src/core/state/State.js
```diff
@@ -3532,18 +3538,28 @@ class State extends ReadyResource {
...
+        const proposedEpochBuffer = proofProposal.epoch;
+        const epochIndexCompare = b4a.compare(nextEpochBuffer, proposedEpochBuffer)
 
-        const currentEpoch = currentEpochBuffer.readBigUInt64BE(0);
-        const nextEpoch = currentEpoch + 1n;
-        const proposedEpoch = proofProposal.epoch.readBigUInt64BE(0);
+        // String helpers
+        const currentEpochStr = toUIntString(currentEpochBuffer);
+        const nextEpochStr = toUIntString(nextEpochBuffer);
+        const proposedEpochStr = toUIntString(proposedEpochBuffer);
```
leonardotc: We are creating this just to safeLogApply that there is a mismatch? Is this data even somehow valuable to the user?

src/core/consensus/services/EpochCoordinatorService.js
```diff
@@ -344,7 +344,26 @@ class EpochCoordinatorService extends SchedulableService {
...
+                        !/^\d+$/.test(epoch)
+                    ) {
+                        this.#logger.error(
+                            '[EpochCoordinatorService] Ignoring EPOCH_CREATED: epoch must be a decimal uint64 string.'
+                        );
+                        return;
+                    }
+
+                    const epochUInt = BigInt(epoch);
+                    if (epochUInt > 0xFFFFFFFFFFFFFFFFn) {
```
leonardotc: Validation doesnt make sense in this context. Epoch created is a success report, its reasonable to accept that it can be expressed as a bigint.

src/core/state/State.js
```diff
@@ -3558,21 +3574,21 @@ class State extends ReadyResource {
...
-        const currentConsensusConfigBuffer = await this.#getEntryApply(EntryType.CONSENSUS_CONFIG_CURRENT, batch);
-        if (currentConsensusConfigBuffer === null) {
+        const currentConsensusConfigIndexBuffer = await this.#getEntryApply(EntryType.CONSENSUS_CONFIG_CURRENT, batch);
+        if (currentConsensusConfigIndexBuffer === null) {
             this.#safeLogApply(OperationType.SET_EPOCH, "Consensus config is not initialized.", node.from.key)
             return Status.FAILURE;
         }
 
-        const currentConsensusConfigIndex = safeReadUint32BE(currentConsensusConfigBuffer);
+        const currentConsensusConfigIndex = safeReadUint32BE(currentConsensusConfigIndexBuffer, 0);
```
leonardotc: You can remove the zero, i think. Since this is an offset, the default actually make sense as a vanilla conversion. Also, is this being used for something? If you just want to validate that it can be represented as uint, doesnt it make sense to validate the length?

src/core/state/State.js
leonardostsouza: - `currentEpochStr ` is used to fetch data from ledger here:
```js
const currentEpochHash = await this.#getEntryApply(EntryType.EPOCH + currentEpochStr, batch);
if (currentEpochHash === null || !b4a.equals(currentEpochHash, proofProposal.previous_epoch_record_hash)) {
    this.#safeLogApply(OperationType.SET_EPOCH, `Previous epoch record hash mismatch for epoch ${currentEpochStr}.`, node.from.key)
    return Status.FAILURE;
}
```
- `nextEpochStr` is used when we write the proof hash in the batch, to be written on ledger:
```js
await batch.put(EntryType.EPOCH + nextEpochStr, epochProofHash);
```
- `proposedEpochStr` was used in a previous code version, but now is really just used for logging. I am going to remove it.

src/core/state/State.js
leonardostsouza: It is possible to use `addBuffers` from `balance.js`, but this would imply exporting this function from that file (it is now only local) and using it here. 

Is this really the best way to proceed? Seems a bit off, unless we create a follow-up refactor task.

src/core/consensus/services/EpochCoordinatorService.js
leonardostsouza: You are right. 
Also, I'm going to change the input for this listener from string to buffer. Seems cleaner

src/core/consensus/services/EpochCoordinatorService.js
leonardostsouza: Done

src/core/consensus/services/EpochCoordinatorService.js
leonardostsouza: Done

src/core/state/State.js
leonardostsouza: Done

src/utils/buffer.js
leonardostsouza: Done

src/core/state/State.js
```diff
@@ -3532,18 +3539,31 @@ class State extends ReadyResource {
...
-        const proposedEpoch = proofProposal.epoch.readBigUInt64BE(0);
+        const proposedEpochBuffer = proofProposal.epoch;
+        const epochIndexCompare = b4a.compare(nextEpochBuffer, proposedEpochBuffer)
 
-        if (proposedEpoch < nextEpoch) {
-            this.#safeLogApply(OperationType.SET_EPOCH, `Stale epoch proposal. Epoch ${currentEpoch} is already committed.`, node.from.key)
+        // String helpers
+        const currentEpochStr = safeToUIntString(currentEpochBuffer);
+        const nextEpochStr = safeToUIntString(nextEpochBuffer);
+        if (nextEpochStr === null || proposedEpochStr === null) {
```
b-garbacz: proposedEpochStr is not defined, you check the variable that does not exist

src/core/state/State.js
```diff
@@ -3558,24 +3578,24 @@ class State extends ReadyResource {
             return Status.FAILURE;
         }
 
-        const currentEpochHash = await this.#getEntryApply(EntryType.EPOCH + currentEpoch.toString(), batch);
+        const currentEpochHash = await this.#getEntryApply(EntryType.EPOCH + currentEpochStr, batch);
         if (currentEpochHash === null || !b4a.equals(currentEpochHash, proofProposal.previous_epoch_record_hash)) {
```
b-garbacz, line 3596: Since we compare the hash of the current active epoch, which will soon become the previous epoch hash, should this field be renamed differently? Not now, maybe in another task where we want to fix structure to be more generic

src/core/state/State.js
```diff
@@ -3558,24 +3578,24 @@ class State extends ReadyResource {
...
+        const currentConsensusConfigIndexBuffer = await this.#getEntryApply(EntryType.CONSENSUS_CONFIG_CURRENT, batch);
+        if (currentConsensusConfigIndexBuffer === null) {
             this.#safeLogApply(OperationType.SET_EPOCH, "Consensus config is not initialized.", node.from.key)
             return Status.FAILURE;
         }
 
-        const currentConsensusConfigIndex = safeReadUint32BE(currentConsensusConfigBuffer);
-        if (currentConsensusConfigIndex === null) {
-            this.#safeLogApply(OperationType.SET_EPOCH, "Failed to read current consensus config index from buffer", node.from.key)
+        if (currentConsensusConfigIndexBuffer.length !== CONSENSUS_CONFIG_INDEX_SIZE) {
```
b-garbacz, line 3607: is it not to defensive? We can keep it but it points that config is broken and time generation won't work anymore.

src/core/state/State.js
```diff
@@ -3515,7 +3522,7 @@ class State extends ReadyResource {
         return Status.SUCCESS;
     }
 
-    async #handleApplySetEpochOperation(op, view, base, node, batch) {
+    async #handleApplySetEpochOperation(op, _view, base, node, batch) {
         if (!this.#stateValidationSchema.validateSetEpochOperation(op)) {
             this.#safeLogApply(OperationType.SET_EPOCH, "Contract schema validation failed.", node.from.key)
```
b-garbacz, line 3527: Previously, we validated the requester address before accepting the operation. Even though indexers are trusted, should we keep a similar check here to reject malformed, testnet, or dev addresses before they can be propagated through the network?
```
        const requesterAddressBuffer = op.address;
        const requesterAddressString = addressUtils.bufferToAddress(requesterAddressBuffer, this.#config.addressPrefix);
        if (requesterAddressString === null) {
            this.#safeLogApply(OperationType.BOOTSTRAP_DEPLOYMENT, "Requester address is invalid.", node.from.key)
            return Status.FAILURE;
        };

        // validate requester public key
        const requesterPublicKey = tracCryptoApi.address.decodeSafe(requesterAddressString);
        if (b4a.equals(requesterPublicKey, NULL_BUFFER)) {
            this.#safeLogApply(OperationType.BOOTSTRAP_DEPLOYMENT, "Failed to decode requester public key.", node.from.key)
            return Status.FAILURE;
        };
```

src/core/state/State.js
b-garbacz, line 3638: Regarding to this:
https://github.com/Trac-Systems/main_settlement_bus/pull/780/changes#r3862562588

We can compare if proposer address from proofData is exactly the same like from `op.address`.

src/core/state/State.js
```diff
@@ -3607,35 +3631,9 @@ class State extends ReadyResource {
...
-            return Status.FAILURE;
-        }
-
+        const indexers = Object.values(base.system.indexers);
         const indexerAddresses = new Set();
-        for (const indexer of Object.values(base.system.indexers)) {
+        for (const indexer of indexers) {
             const indexerAddressBuffer = await this.#getRegisteredWriterKeyApply(batch, indexer.key.toString('hex'));
             if (!indexerAddressBuffer) continue;
             const indexerAddress = addressUtils.bufferToAddress(indexerAddressBuffer, this.#config.addressPrefix);
```
b-garbacz, line 3653: In your opinion does it worth to add check here if indexerAddress === null then return Failure?

src/core/state/State.js
```diff
@@ -3648,6 +3646,25 @@ class State extends ReadyResource {
             return Status.FAILURE;
         }
 
+        const indexerCount = indexers.length;
+        const quorumThreshold = indexerCount <= 2 ? 1 : Math.floor(indexerCount / 2) + 1;
+        const approvalCount = 1 + op.seo.app.length; // proposer plus submitted approvals (proposer's own verified signature counts as one approval)
```
b-garbacz, line 3665: My suggestion is to count approvals only after verifying them. What do you think?

If even one approval is malformed or invalid, should we reject the whole proof? Or should we delete the invalid approval and continue with the remaining valid ones?

src/core/state/State.js
b-garbacz: Do we need this try catch if we have `blake3Safe`?

src/core/state/State.js
```diff
@@ -3702,21 +3716,64 @@ class State extends ReadyResource {
...
 
+    async #validateEpochApprovals(encodedApprovals, indexerAddresses, proposerAddress, challengeData, proofProposal) {
+        const approverIdentities = new Set();
+
+        for (const encodedApproval of encodedApprovals) {
+            const approval = safeDecodeProofProposalApproval(encodedApproval);
+            if (approval === null) return 'Failed to decode epoch approval.';
+
+            const approverAddress = addressUtils.bufferToAddress(approval.approver, this.#config.addressPrefix);
+            if (!approverAddress) return 'Epoch approval contains an invalid approver.';
```
b-garbacz, line 3754: Question - can not we use `this.#safeLogApply` and return just null? (I am referring to alll errors)

src/core/state/State.js
```diff
@@ -3702,21 +3716,64 @@ class State extends ReadyResource {
...
+            if (approverIdentities.has(approverAddress)) return 'Epoch approval contains a duplicate approver.';
+
+            const approverPublicKey = tracCryptoApi.address.decodeSafe(approverAddress);
+            if (b4a.equals(approverPublicKey, NULL_BUFFER)) return 'Failed to decode epoch approver public key.';
+
+            const approvalMessage = createMessage(
+                challengeData,
+                proofProposal.proof,
+                approval.approver,
+                proofProposal.signature
```
b-garbacz, line 3766: We can create P2/P3 task to move this approver on the bottom of the byte order

src/core/state/State.js
```diff
@@ -3702,21 +3716,64 @@ class State extends ReadyResource {
...
+            );
+            let approvalVerified = false;
+            try {
+                const approvalHash = await tracCryptoApi.hash.blake3(approvalMessage);
```
b-garbacz: safeBlake

b-garbacz: then we don't need this try catch I think

src/core/state/State.js
```diff
@@ -3702,21 +3716,64 @@ class State extends ReadyResource {
...
+                    approval.approval_sig,
+                    approvalHash,
+                    approverPublicKey
+                );
+            } catch {
+                approvalVerified = false;
+            }
+            if (!approvalVerified) return 'Failed to verify epoch approval signature.';
+
+            approverIdentities.add(approverAddress);
```
b-garbacz, line 3778: don't we want to return this and use it somehow?

src/core/state/State.js
b-garbacz: However...It's fine tbh

src/core/state/State.js
leonardostsouza: Sure. This is important to have

src/core/state/State.js
leonardostsouza: This is checking for local storage corruption, not config. I agree it may be overkill, however

src/core/state/State.js
leonardostsouza: Fixed

src/core/state/State.js
leonardostsouza: We can, but this is kinda useless. The `op.address` value is never signed and this is by design. There would be no problem to relay the epoch proof through some other node in case it is necessary in the future. I decided to drop this check.

src/core/state/State.js
leonardostsouza: In my opinion this is overkill. It one of the indexer records is corrupted, and this indexer is not the proposer, we can just continue here.

src/core/state/State.js
leonardostsouza: If even one approval is invalid or malformed, then we can assume the proposer is compromised, because it should have checked this thing. We should invalidate the whole proposal and wait for a well formed one.

src/core/state/State.js
leonardostsouza: No. Replaced

src/core/state/State.js
leonardostsouza: No, because the errors might be in several places. The strings are the actual error message we are going to show in the safeLogApply

src/core/state/State.js
leonardostsouza: Fixed

src/core/state/State.js
leonardostsouza: This is leftover from a previous refactor that I forgot to remove. Fixed now

leonardostsouza: No, this is only being used internally to check if there are duplicate approvers. In the end, all approvals should be valid or the whole proposal is discarded

b-garbacz (approved):

Issue #765, opened 2026-08-06
feat: emit event when consensus config parameters are changed


leonardostsouza:
Objectives:
- Emit the event when the relevant apply operation is called
- handle the event by stopping ongoing vdf calculation and restarting with new parameters

PR #785, opened 2026-08-20
feat: emit event then consensus config parameters are changed (#765)

b-garbacz:


src/core/consensus/services/EpochCoordinatorService.js
```diff
@@ -20,348 +11,288 @@ class EpochCoordinatorService extends SchedulableService {
...
-            if (next === EPOCH_STATES.COLLECT_APPROVALS && prev !== EPOCH_STATES.COLLECT_APPROVALS) {
-                signatureTimer = setTimeout(() => stateMachine.send(EPOCH_EVENTS.APPROVAL_COLLECTION_FAILED), this.#config.epochSignatureTimeout);
-            } else if (prev === EPOCH_STATES.COLLECT_APPROVALS && next !== EPOCH_STATES.COLLECT_APPROVALS) {
-                clearTimeout(signatureTimer);
-            }
+            await schedulerStopped;
+            await roundCancelled;
+            await vdfReload;
+        } catch (error) {
+            this.#logger.error(`[EpochCoordinatorService] config reload failed: ${error.message}`);
```
jusufsuljic: If this happens is this fatal or are we expecting to run a retry or something?

src/core/state/State.js
```diff
@@ -439,6 +448,19 @@ class State extends ReadyResource {
...
 
+    /**
+     * Publishes apply events after the related state changes become signed.
+     * The optional length is used by tests which provide their own Autobase instance.
+     *
+     * @param {number} signedLength current signed view length
+     */
+    handleViewUpdate(signedLength = this.getSignedLength()) {
+        this.#stateEventQueue.flush(signedLength, (event, ...args) => {
+            console.log('[ConsensusReloadDebug] dispatching published state event', event);
```
jusufsuljic: We are keeping the debug logs?

src/core/consensus/services/EpochCoordinatorService.js
b-garbacz: Good observation. This would be good to have a retry

jusufsuljic (approved):

